PR #489, opened 2026-09-08
feat(worktrees): reserve Change Stream worktree locations (#211)

RepairYourTech:
Closes toward #211 (location layer only; the issue remains open for Git materialization and the rest of its acceptance). Respects the #190 repository abstraction boundary: this crate never runs Git.

## What this adds

**`symbiote-worktrees`** — deterministic collision-resistant naming plus filesystem reservation and re-verification of the location the Linux sandbox requires its caller to reserve (the trusted-provisioner premise documented in `docs/security/linux-sandbox.md`).

- **Deterministic naming.** `Derived::derive` is a pure function of canonical Project/Root/Change Stream identities plus a trusted policy seed (1–64 chars `[A-Za-z0-9_-]`, from Host state, never client JSON). It yields a bounded `st-…` WorktreeId string and a Git-ref-safe branch `symbiote/<project>/<stream>/<12hex>` (every component check-ref-format validated; no leading dot, no `.lock`, no `..`, no `@{`). Same recorded inputs → same names; different identities or seeds never collide. The derived names feed the store's existing `create_task` worktree/branch uniqueness enforcement.
- **Reservation.** `reserve()` creates `<base>/<project>/<worktree_id>/` with mode-0700 private directories and an `O_EXCL` fsynced marker binding the derived identity, so concurrent reservations have exactly one winner. Premises mirror the sandbox launcher: bounded absolute symlink-free paths, system trees (`/usr`, `/etc`, `/dev`, `/proc`, `/sys`, `/run`, `/boot`, `/var`) refused, euid/root ownership with the sandbox's exact sticky-bit walk rule, failed marker writes clean up their directory.
- **Verification.** `verify()` creates nothing: it re-walks without following symlinks, checks ownership/modes, reads the marker, rejects identity mismatch and any foreign content in the reserved directory. The sandbox launcher still revalidates independently; this is the provisioner side of the documented premise.
- **Release discipline.** `release()` has exactly two outcomes: `Retain` (marker removed, directory kept as an abandoned location) or `DeleteIfEmpty` (only when the directory holds zero entries). Any content — materialized Git state or uncommitted files — fails `NotEmpty` and touches nothing. Unmerged work is never deleted without evidence and authorized confirmation, per the Change Stream contract.

The marker lives in the private project directory, not inside the worktree, so later `git worktree add` materialization keeps a clean tree.

## Honest non-claims

[...]

Issue #94, opened 2026-09-03
[WORK-02][Wave 1] Build the canonical task DAG, dependencies, ownership, foreign runtime links, and state machine


RepairYourTech:
**Canonical implementation issue.** Parent epic: #159 · Wave: 1 · Audit: #443.

## Accepted v2.4 requirements and corrections

The following incorporates the latest accepted decisions and controls interpretation of retained baseline wording.

### Native workforce, goals and local-delegation amendment
- Permit selected Lead and every worker Role to use Symbiote Agent or an external harness independently. No external CLI or proprietary model is mandatory for native-only operation.
- Bind runtime kind/profile, model provider/account, authorized billing entitlement, root effort, allowed local-child profiles/efforts, goal strategy, context, permissions and total resource limits explicitly. Use #460/#461/#462 contracts.
- Preserve canonical Task/Capability/Change Stream and runtime-local child distinctions. Child agents report to the parent dispatch; independent discovered work is promoted via normal task/issue policy, not automatically duplicated in GitHub.
- Root plans, partitions, integrates and validates children; final independent review remains a separate configured gate. Root/child native completion is evidence only.
- Host preflight prepares read-only config/access/context plans, authorizes and persists the contract, establishes enforcement, then activates runtime resources. No executable hooks/extensions run before this boundary.
- Goal pause/resume/cancel is durable, budget-preserving and fenced; no-progress/failed-verification/unknown-side-effect states are explicit. Project switching does not kill leased background work.
- Instrument complete root/child/retry/fallback episodes and include analytics completeness. Stop known-unverifiable hard limits before launch rather than pretend enforcement.
- Learning can recommend changes or optimize within preapproved pools only. Explicit Role/model assignments and paid fallback consent cannot be silently changed.
- [ ] External-only/native-only/hybrid fixtures preserve the same authority and handoff semantics.
- [ ] All descendants remain within aggregate depth/concurrency/budget/permission boundaries or the runtime is marked insufficient for that contract.
- [ ] Deterministic gates remain active when learning, analytics upload or external CLIs are unavailable.

## Outcome
Give Symbiote authoritative deterministic control of engineering work independent of any harness's internal todo list or transcript, while cleanly separating canonical work semantics from Git mutation/integration streams.

## Required scope
- Typed dependencies: blocks, requires, consumes-contract-from, reviews, verifies, supersedes, conflicts-with, follow-up-to.
[...]

PR #490, opened 2026-09-08
feat(work): add typed Task dependency edges with DAG enforcement (#94)

RepairYourTech:
Closes toward #94 (first dependency-ready slice; the issue remains open for the state machine, foreign runtime links, and DAG queries). Advances #189's task-schema scope too. Depends on the #189/#43 foundations already on main.

## What this adds

**Typed dependency edges (`symbiote-domain::dependency`).** Eight kinds (`requires`, `consumes_contract_from`, `blocks`, `reviews`, `verifies`, `supersedes`, `conflicts_with`, `follow_up_to`) with cross-Project-capable targets. Edges are separate first-class records owned by one Task with whole-set replacement semantics — deliberately not Task fields or lifecycle commands, since a Task's pure `apply` cannot validate a graph spanning aggregates. The global validator treats every edge as one graph across Projects, inverts the waits-on direction for `blocks`, and rejects self-edges, dangling targets and cycles via Kahn's algorithm (≤16,384 tasks, ≤64 edges/task).

**Durable edges + completion gate (`symbiote-store` schema v7).** A `task_dependencies` table with tamper-evident indexed columns (same convention as bindings/teams/routes). `set_task_dependencies` re-validates the entire DAG inside the write transaction with exact idempotency; journal replay re-derives each set and re-validates the graph at every journal position. `apply_task` now refuses completion while any `requires`/`consumes_contract_from`/incoming-`blocks` target is not `Completed` — a typed error, journal untouched. Non-blocking kinds are recorded with provenance and **documented as not yet enforced** (their policy needs review/merge/supersession states that don't exist yet).

**Protocol v1.7.** `set_task_dependencies` (`ManageWork` on the owning Project + `Read` on every target Project, mirroring the cross-Project work rule), `get_task_dependencies` (read), the `task_dependencies_set` journal payload, and `TaskDependencyWrite`/`TaskDependencyRead` capabilities. The Task-draft branch bound rises to 512 bytes to match the #489 derivation bound. All fixtures bumped to v1.7.

**End-to-end:** daemon test drives set → restart replay → readback → cycle rejection (typed `conflict`, host keeps serving) → self-edge `invalid_request` → non-blocking kind recorded → journal lineage.

## Honest non-claims

No foreign runtime links, no DAG readiness/critical-path/blocker queries, no blocked/queued/assigned/leased task states, no multi-task Change Streams, no stream-collision blocking of member tasks. The completion gate is one enforced policy point, not the full #94 acceptance. #94 stays open.

## Verification

- `cargo test --workspace`: 294 passed, 0 failed (6 new domain graph tests, 4 new store tests incl. tamper/cycle/replay/restart, extended daemon flow, migration tests extended to v7)

## Review scope

DAG direction semantics (the classic double-inversion bug around `blocks`), replay-audit symmetry, cross-Project authorization, and the honest enforced-vs-recorded split are what I most want independent eyes on.


RepairYourTech:

Two full review passes plus Qodo re-reviews after every fix commit:

1. **Independent reviewer agent** (full diff at `e0088237`, all five focus areas): 3 P0 + 1 P1 + 4 P2. The three P0s — the `blocks` double-inversion in the completion gate, the empty-set write that made the store unopenable on restart (reviewer confirmed by execution), and the `lineage_matches` rejection of legal cross-project targets — are all fixed in `25152ff6` with regression tests. The P1 (journal/read disclosure of cross-project identifiers) is fixed by requiring `Read` on every referenced target project in both the journal stream and the read endpoint, matching the work-reference precedent. P2s: tautological kind check removed, unused audit parameter dropped, replay-strength claim corrected in the PR evidence below.
2. **Qodo**, re-triggered at `25152ff6` and `e751b13`: its re-review confirmed the earlier fixes by not re-raising them and surfaced two new provenance findings — the audit arm ignoring the payload's declared project, and injected table rows with no journal provenance evading the audit. Both fixed in `e751b13`: the declared project must equal the journal row's project, and every materialized owner must have journal provenance.
3. **CI**: green at `e751b13` (MSRV 1.85 + stable contracts, host-lifetime, offline-validation, shell-client-build). Local: 295 workspace tests passing, fmt and strict clippy clean.

Known limits, stated rather than papered over: replay validates the dependency graph once at end-of-journal while the write path validates per transaction (distinguishing per-position corruption from end-state corruption requires ordered per-event validation, deferred with the #94 state machine slice); a `Cancelled`/`Failed` blocking target releases only by replacing the owner's edge set (operational caveat, documented); `Completed` tasks can acquire new blocking edges after the fact — future hardening will reject edge sets that gate already-completed work retroactively.

Issue #95, opened 2026-09-03
[WORK-03][Wave 2] Implement the event bus, scheduler, leases, wake-up orchestration, retries, and approval waits


RepairYourTech:
**Canonical implementation issue.** Parent epic: #159 · Wave: 2 · Audit: #443.

## Accepted v2.4 requirements and corrections

The following incorporates the latest accepted decisions and controls interpretation of retained baseline wording.

### Native workforce, goals and local-delegation amendment
- Permit selected Lead and every worker Role to use Symbiote Agent or an external harness independently. No external CLI or proprietary model is mandatory for native-only operation.
- Bind runtime kind/profile, model provider/account, authorized billing entitlement, root effort, allowed local-child profiles/efforts, goal strategy, context, permissions and total resource limits explicitly. Use #460/#461/#462 contracts.
- Preserve canonical Task/Capability/Change Stream and runtime-local child distinctions. Child agents report to the parent dispatch; independent discovered work is promoted via normal task/issue policy, not automatically duplicated in GitHub.
- Root plans, partitions, integrates and validates children; final independent review remains a separate configured gate. Root/child native completion is evidence only.
- Host preflight prepares read-only config/access/context plans, authorizes and persists the contract, establishes enforcement, then activates runtime resources. No executable hooks/extensions run before this boundary.
- Goal pause/resume/cancel is durable, budget-preserving and fenced; no-progress/failed-verification/unknown-side-effect states are explicit. Project switching does not kill leased background work.
- Instrument complete root/child/retry/fallback episodes and include analytics completeness. Stop known-unverifiable hard limits before launch rather than pretend enforcement.
- Learning can recommend changes or optimize within preapproved pools only. Explicit Role/model assignments and paid fallback consent cannot be silently changed.
- [ ] External-only/native-only/hybrid fixtures preserve the same authority and handoff semantics.
- [ ] All descendants remain within aggregate depth/concurrency/budget/permission boundaries or the runtime is marked insufficient for that contract.
- [ ] Deterministic gates remain active when learning, analytics upload or external CLIs are unavailable.

## Outcome
Coordinate agents and Change Streams through durable events and scheduling rather than wasting model context on polling loops, while reacting immediately when diagnostics or collision evidence make active parallel work unsafe.

## Required scope
[...]

PR #491, opened 2026-09-08
feat(sched): add durable task leases with fencing and scheduling projection (#95)

RepairYourTech:
Closes toward #95 (lease/fencing + explainable scheduling slice; the issue remains open for event wake/notify, collision consumption, dead-letter delivery, and budget/fabric ownership). Depends on #94's dependency edges (merged in #490), #181's protocol and the #210 journal provenance foundations.

## What this adds

**Durable leases with fencing tokens (`symbiote-domain::lease` + store schema v8).** A `TaskLease` binds one `Running` task's dispatch to a host for a bounded window (1s–1h). Fencing tokens are monotonic per task: post-expiry acquisition takes the next generation, a superseded generation is fenced out of release, and renewal by the same holder extends the window without burning a token. Transitions are `held → expired | released | fenced`; only an unexpired `held` lease holds.

- `acquire_lease` validates the task is `Running` and the dispatch/host match the stored current dispatch — a foreign dispatch is `invalid_request`, never a silent re-staff.
- `expire_stale_leases` is an idempotent journaled sweep under deterministic per-task command IDs. Expiry **never mutates task state on its own**: the Task lifecycle (Interrupt/Recover) stays the Host's decision, keeping worker completion separate from verified completion.
- Journal replay (v8) verifies lease shape, project lineage (payload `project_id` must equal the journal row), request bytes, and the replay-time dispatch invariants; the table must replay exactly and injected rows with no journal provenance fail startup.
- Tamper-evident indexed columns follow the binding/dependency/route convention; migration tests extended to v8.

**Explainable scheduling projection.** `scheduling_projection` classifies every `Ready` task with a recorded reason — `stream_unsafe` (Change Stream not `Active`) before `stream_leased` (live lease on the same stream serializes by policy) before `dependency_unresolved` (outgoing `requires`/`consumes_contract_from` target not `Completed`); schedulable tasks carry `no_blocking_dependencies` or `dependencies_satisfied`. Outgoing `blocks` deliberately does not gate starting — it gates the target's completion, matching #94's direction semantics.

**Protocol v1.8.** `acquire_task_lease`/`release_task_lease` (Host-local authority until #269 introduces restricted worker identities), `expire_stale_leases`/`get_scheduling_projection` (owner), `scheduler_sweep` response, `task_leased` journal payload, `TaskLeaseManagement`/`SchedulingProjection` capabilities. Fixtures bumped to v1.8.

## Honest non-claims

Leases authorize nothing by themselves — dispatch contracts, permissions and consent stay upstream. Pending #95 acceptance: event-driven wake/notify (idle Lead consumes no inference), duplicate completion/collision deduplication, collision-evidence consumption to pause/serialize streams, dead-letter visibility, budget/concurrency/Host-resource accounting, fabric-distributed dispatch ownership, and automated reassignment after stale-lease recovery.

## Verification

- `cargo test --workspace`: 301 passed, 0 failed (2 domain lease tests, 3 store lease tests covering acquire/renew/expiry-idempotence/token-generations/fenced-release/foreign-dispatch/not-Running/restart-replay, daemon projection/sweep test, migration tests extended to v8)

## Review scope

Fencing semantics (token generations vs release permission), the expiry-sweep's deliberate non-mutation of task state, replay-time lease invariants, and the projection's precedence order are what I most want independent eyes on.


RepairYourTech:

Two full independent passes plus four Qodo rounds:

1. **Independent reviewer agent** (full diff at `37d79f3f`, five focus areas with adversarial probes): the fencing/token core, sweep mechanics, authorization, projection precedence, and the #490-class inverse audit were all verified sound. It found one P0 — terminal lease events journaled after the task left `Running` bricked replay on the documented interrupt-then-sweep recovery path (probe-confirmed) — plus journal-page and sweep-ID poisoning P1s and five P2s. All fixed in `25c78d0e` with a regression test for the exact recovery sequence, the replay state machine with strict token progression, reserved sweep IDs, retry-before-refusal release, authenticated attribution, error propagation, window-bounded shape validation, and table-vs-journal count reconciliation.
2. **Qodo** re-reviews at each head. Round 2 confirmed the earlier fixes by not re-raising them and caught three real issues in my fixes: post-terminal acquisition must be allowed to rebind dispatch/host (recovered task under replacement dispatch), the reserved namespace must actually cover the sweep IDs the code generates, and token+1 must be checked arithmetic. All three fixed in `764bec7a`; its remaining round-2 items (unbounded projection scan, `no_blocking_dependencies` reason accuracy, projection trusting indexed columns without body cross-check) are explicitly deferred as documented scope, listed below.
3. **CI**: green at every pushed head including final `764bec7` (MSRV 1.85 + stable contracts, host-lifetime, offline-validation, shell-client-build). Local at final head: 302 workspace tests passing, fmt clean, strict clippy clean.

**Known limits, stated rather than papered over** (tracked under #95): the scheduling projection is unbounded in one pass (task counts bounded by store caps, response size enforced downstream; pagination deferred); `no_blocking_dependencies` currently means "no *gating* edges" and mislabels tasks with only non-gating edges; the projection reads indexed columns without re-validating against bodies (the write path and reopen audit do); replay of lease events validates state transitions and token progression but not per-event target existence ordering; event wake/notify, collision consumption, dead-letter delivery, budget accounting, and automated reassignment after stale-lease recovery remain the bulk of #95.

Issue #36, opened 2026-09-03
[FND-03][Wave 0] Define the canonical domain, security, event, and lifecycle model


RepairYourTech:
## Accepted technology direction — 2026-09-08 takeover

Rust is the core implementation direction. Keep protocol identities independent of language/process. Desktop is a Host client; CLI and desktop share native execution. Process isolation follows trust/recovery/resource needs. #170 and #38 govern accepted technology and pending proof.

**Canonical implementation issue.** Parent epic: #155 · Wave: 0 · Audit: #443.

## Accepted v2.4 requirements and corrections

The following incorporates the latest accepted decisions and controls interpretation of retained baseline wording.

### Accepted architecture additions and precedence
- Support EXTERNAL_HARNESS and NATIVE_SYMBIOTE runtimes as peers: external-only, hybrid and fully native Teams preserve the same canonical Project, Role, Task, Change Stream, knowledge, policy and completion engine. Symbiote Agent is first-party execution software using supported models, not a required proprietary model.
- Separate AgentRuntimeAdapter from InferenceProviderAdapter, ProviderConnection, CredentialReference and BillingEntitlement. Embedding a third-party full agent loop remains an external runtime, even when rendered in our UI.
- Native Symbiote CLI, desktop and remote clients use one shared agent implementation/Host; fully native operation requires no Claude Code/Codex/Pi/OpenCode installation. Existing headless administrative CLI scope is retained.
- Adopt bounded goal execution and intra-harness delegation: independently configured root/child models and effort where verified, explicit local-child lineage, aggregate limits and cancellation; native children do not silently become global Roles or bypass staffing/permissions.
- Add opt-in scoped project/client-method learning, controlled experiments, and local global/Project/task analytics. Learning cannot override security, verified entitlement, billing consent, fixed staffing, locked product decisions or evaluation gates.
- Every efficiency/improvement claim needs comparable outcome evidence, denominator/sample/missingness and full root/child/review/remediation cost. Unknown model attention, hypothetical defects prevented and counterfactual savings cannot be presented as measurements.
- Source is evidence of current behavior; approved specifications are authority for intended behavior. A contradiction is recorded and remediated, not automatically resolved by rewriting docs to match faulty code.
- The normative engineering target is no known unresolved release-blocking defect within the verified scope, not a universal zero-defect or perfect-code guarantee. Approved debt, unresolved dynamic boundaries and unmeasured behavior remain visible.

### Additional acceptance
[...]

PR #492, opened 2026-09-08
feat(prov): persist provider, entitlement, and model identity records (#464)

RepairYourTech:
Closes toward #464 (durable registry slice; the issue remains open for dispatch/session flow, mixed-runtime round-trips, and the adapter/session persistence layer). Depends on #184's SDK provider contracts and the #36 foundations on main.

## What this adds

**Durable provider identity (`symbiote-store` schema v9).** The #184 SDK contracts (`symbiote-runtime-sdk::provider`) already define `ProviderConnection`, `BillingEntitlement`, `ModelDescriptor`, `ProviderBinding` and the request/response envelopes as types. This slice makes the identity records durable, replay-audited state:

- `provider_connections` (adapter + endpoint + authentication kind, replacements supersede with journal revision history), `billing_entitlements` (FK to a stored connection; expiry validated at SDK binding time, not registration), `model_descriptors` (SDK-validated at registration: schema version, context/output bounds, capabilities).
- Tamper-evident indexed columns following the binding/dependency/route/lease convention. Journal replay validates shape, attribution, request-byte equality, connection precedence, and reconciles all three tables against the journal in both directions — the #490/#491 lesson applied from the start.
- Registration is **attributed** to a real Project the caller names (journal rows are project-scoped, so provenance is explicit) while records stay global identity, re-authorized at binding time. Attribution scopes nothing.

**Protocol v1.9.** `replace_provider_connection` / `replace_billing_entitlement` / `replace_model_descriptor` (local-owner authority — restricted worker identities are #269's boundary; #217 keeps secret materialization out of this registry), three owner-only reads, three journal payloads, `ProviderRegistryWrite`/`Read` capabilities. Fixtures bumped to v1.9.

## Honest non-claims

No dispatch/session flow, no mixed-runtime schema round-trips (external Lead/native worker etc. — needs the dispatch layer), no missing-CLI or local-endpoint-only startup tests, no AgentRuntimeAdapter/session persistence. #464 stays open; these are the next slices.

## Verification

- `cargo test --workspace`: 304 passed, 0 failed (2 new store tests: register/replay/restart/replacement/unknown-connection/invalid-descriptor; tampered-row reopen refusal; migration tests extended to v9)

## Review scope

The attribution-vs-global-identity split (project-scoped journal rows for global records), replay symmetry, and the owner-authority boundary are what I most want independent eyes on.


RepairYourTech:

Two review channels, two rounds each:

1. **Independent reviewer agent** (full diff at `e1e02047`, four focus areas): no P0. Verified the event-sourcing mechanics symmetric (replacement last-journal-wins, forward-reference enforcement on both sides, byte-equality via canonical re-serialization, complete payload mapping) and the honest-labeling claims true. Found 2 P1 + 5 P2:
   - P1 authorization coverage → `provider_registry_writes_and_reads_stay_owner_authority_only` in `crates/symbiote-protocol/tests/contracts.rs` (restricted principal with full project grants denied on all six operations; local_owner allowed).
   - P1 attribution/journal doc-vs-behavior → the interlocking journal filter landed in `b67078a3` (registry payload bodies withheld from non-owner journal reads) and the contract doc now states exactly that; the descriptor-validation claim is trimmed to what `ModelDescriptor::validate` checks.
   - P2s: the vacuous attribution Read conjunct is removed; journal replay classifies an invalid descriptor as `Integrity`; lineage validation mirrors shape checks for entitlements; the audit error names the specific table.
2. **Qodo**, two rounds at `e1e02047`: six findings (lineage strictness for model events and timestamps; row-key/body identity binding on all three readers; Revision(0) enforcement in replay; attribution-project precedence in replay; owner-only filtering of registry payloads in project journal reads). All fixed in `b67078a`.
3. **CI**: green at every pushed head including final `9ebfe5e`. Local: 305 workspace tests passing, fmt clean, strict clippy clean. Qodo's `/review` at the final head did not produce a fresh comment within the wait window; the disposition above covers every finding it raised, each with a fix commit.

PR #493, opened 2026-09-08
feat(disp): add durable dispatch preparation composing the merged primitives (#205)

RepairYourTech:
Closes toward #205 (dispatch composition slice; the issue remains open for Start-transition consumption, worker loops, and the least-privilege manifest). Composes the five merged primitives: Role routing (#204/#488), dependency edges (#94/#490), task leases (#95/#491), worktree locations (#211/#489), and the provider registry (#464/#492).

## What this adds

**`prepare_dispatch(task_id)`** — a durable composition record built entirely from authoritative storage:

- **Scheduling**: the task must be `Ready` (the #95 projection's core condition).
- **Routing**: the latest route decision (#204) for the task's *origin work identity* supplies the resolved Role; unresolved routes are recorded as such.
- **Lease**: a `held`, unexpired lease (#95) contributes its dispatch binding and fencing token — the token a later `Start` transition must carry.
- **Worktree**: the stream's reserved worktree identity and branch from the Change Stream record (#211).
- **Provider**: the routed Role's workforce binding (#203/#486) carries the runtime profile; its provider connection must exist in the #464 registry for this step to resolve.

The outcome is `ready` only when scheduling, routing, lease, and provider all resolve. **A refused composition is recorded like a ready one**: the refusing step is part of the record, refusals are evidence with exact reasons, and the caller receives `failed_precondition` (new error code) while the durable outcome always lands.

**Durability.** Schema v10: `dispatch_preparations` (latest per task, tamper-evident indexed columns), one `dispatch_prepared` journal event per preparation, replay validating shape/attribution/request-bytes, and two-directional table-vs-journal reconciliation — the established store convention applied from the start.

**Protocol v1.10.** `prepare_dispatch` / `get_dispatch_preparation` (local-owner authority, consistent with the scheduler endpoints; restricted worker identities are #269), the `dispatch_prepared` journal payload, a `DispatchPreparation` capability, and the `FailedPrecondition` error code. Fixtures bumped to v1.10.

**Tests.** Store test: composition steps, refusal recording, replay, readback, unknown-task refusal. Daemon test: projection → prepare → refusal recorded → replay identical → restart replay → journal lineage.

## Honest non-claims

No `Start`-transition consumption yet (the preparation identifies what may start; compiling the live Dispatch contract against Host enforcement claims is the next slice), no worker-loop activation (#465/#464), no mixed-runtime round-trips, no least-privilege manifest (#269). #205 stays open.

## Verification

[...]

PR #494, opened 2026-09-09
feat(disp): start prepared tasks under transactionally minted leases (#205)

RepairYourTech:
Closes toward #205 (Start-transition slice; the issue remains open for worker-loop activation, mixed-runtime round-trips, and the least-privilege manifest). Consumes the #205 composition slice merged at e05e3990.

## What this adds

**`start_prepared_task`** — one transaction consuming a `Ready` `DispatchPreparation`:

1. Re-verifies the routed Role still exists; refuses when a live lease from a previous generation is held (`StillHeld`).
2. Compiles the live `WorkforceRuntimeContract`/`Dispatch` via the domain `Dispatch::compile` — Role record, workforce binding, runtime profile, and the Host record with its enforcement claims.
3. Applies the domain `Start` transition (`Task::apply` does all lineage/actor/contract validation).
4. **Acquires the governing lease in the same transaction**: the fencing token is minted at start (progressing on any recovered generation), journaled as a `task_leased` event under a `<command>-lease` id.

Design correction discovered during implementation: first-time starts can never pre-hold a lease — leases exist to fence *Running* dispatches — so the preparation's lease step is now informational and `ready` is the conjunction of scheduling+routing+provider. Refused preparations, superseded preparations, and second starts all fail closed: the task stays Ready and no lease exists.

**Host service** (protocol v1.11, local-owner authority): mints dispatch/contract ids from the inventory nonce (stable per daemon), builds the Host record with operator-provisioned `HostEnforced` claims (Filesystem/Process/Cancellation/CompletionAuthority, one-hour window), and refuses any host identity that is not its own inventory identity. Response carries `StartedDispatch` (dispatch id, stream, fencing token).

**Tests**: store e2e (route → prepare → start → Running with lease token 1 → post-start prepare records the scheduling refusal → second start refused with all three refusal modes asserted) and daemon (start with foreign host identity denied; refused preparation start fails closed with `failed_precondition`; task remains Ready). Fixtures bumped to v1.11.

## Honest non-claims

The started dispatch is a **control-plane record only** — no runtime process is launched (that is #465's native loop and #464/#465's external loop, the direct next slices). Host enforcement claims are operator-provisioned self-assertions, not yet tied to sandbox-observed evidence; the #269 manifest and mixed-runtime round-trips remain open. #205 stays open.

## Verification

- `cargo test --workspace`: 309 passed, 0 failed

## Review scope

The lease-minting-inside-the-Start-transaction semantics, the informational lease step in preparations (conjunction change), the operator-provisioned Host claims boundary, and the failure-closed second-start paths are what I most want independent eyes on.


RepairYourTech:

**Independent reviewer agent** (full diff at `07cfc15`, five focus areas with empirical probes): **1 P0 (confirmed by execution) + 4 P1 + 4 P2** — verdict "do not merge as-is". All fixed in `a04594d0`:

1. **P0** — the Start `TaskChanged` journal row used payload-serialization bytes instead of `mutation_request` bytes, so a committed start failed the audit on reopen and permanently bricked the daemon (the reviewer reproduced this with a temporary test). Fixed to the `apply_task` convention; regression coverage now asserts `integrity_check` passes and the store reopens with the task Running and lease Held after a committed start — the exact scenario the suite was missing.
2. **P1** — idempotent retry of a committed start now replays its recorded receipt (validated by payload kind and task identity, since claim windows move with time and request bytes cannot be re-derived).
3. **P1** — reserved sweep-namespace command ids are refused on start, preserving the anti-poisoning invariant; regression test added.
4. **P1** — Start now enforces the scheduling gates (unresolved completion-blocking dependencies, non-Active streams) plus `apply_task`'s explicit origin guard.
5. **P1** — docs corrected: `ready` is the conjunction of scheduling+routing+provider; the lease step is informational; the fencing token is minted at Start.
6. **P2s** — dispatch/contract ids are stable per task across daemon restarts (the lease audit requires dispatch-id stability across generations); the phantom `dispatch-start.md` reference and the pulse-cadence claim-window wording fixed; the unused nonce accessor removed.

**Qodo**: no bot review was produced on this PR within the observation window (three `/review` triggers) — noted as a coverage gap rather than claimed, consistent with the #493 disposition.

**CI**: green at every pushed head including final `a04594d0` (MSRV 1.85 + stable contracts, host-lifetime, offline-validation, shell-client-build). Local at final head: 309 workspace tests passing, fmt clean, strict clippy clean.

PR #496, opened 2026-09-09
feat(external): add the Codex App Server turn driver with total approval refusal

RepairYourTech:
Closes nothing — this is the external-side slice of #465 (with #464), which stays open.

## What exists now
- New crate `symbiote-external-agent`: drives the pinned Codex App Server (`codex-cli 0.118.0`, established by #483) as an external harness. `ExternalSession` starts a thread (`thread/start`) with the Host-authorized worktree as `cwd`, `sandbox: "read-only"` and `approvalPolicy: "never"`, then files task turns (`turn/start`) and observes them to a terminal status.
- Pinned protocol shapes were verified against the installed executable's `app-server generate-json-schema` output (v2 thread/turn APIs); the versionless JSONL envelope (no `jsonrpc` field, string-or-number ids, notifications interleaved) matches the #483 discovery probe.
- All harness notifications map onto the SDK's normalized event stream with `RuntimeKind::ExternalHarness` session bindings, Ready-first, bounded text (16 KiB, char-boundary truncation with a visible marker). Usage notifications become `Measured` aggregate input/output with billing and cached-input explicitly `Unknown/NotReported` and `coverage: AggregateOnly`; missing numeric fields count as unreported turns — never free.
- **Total approval refusal**: every server-initiated escalation request (`execCommandApproval`, `applyPatchApproval`, `item/permissions/requestApproval`, `item/commandExecution/requestApproval`, `item/fileChange/requestApproval`, `item/tool/requestUserInput`, `mcpServer/elicitation/request`, `item/tool/call`, and any unknown method) is refused, counted, and recorded as a bounded Diagnostic naming only the method — never the payload. The dispatch contract's access snapshot is the only permission authority.
- Credential separation is structural: the driver sends no credentials, no API keys, no native billing route; `account/read` and `refreshToken` are never called. Harness auth stays with the harness account.
- Explicit stop states (`Completed`/`Failed`/`Interrupted`/`TransportLost`); notification budgets bounded; foreign-turn notifications absorbed but never stopping the run; completion reports are RequestCompletion evidence only.

## Verification
- 14 new offline tests (deterministic fixture transports): turn completion/failure/interruption, observations mapping, usage accounting + unreported turns, 5-request refusal matrix incl. unknown-method, transport loss, notification flood bound, native-dispatch refusal (`NotExternalHarness`), expired contract + bad inputs, oversized-text truncation, double-filing refusal.
[...]

PR #483, opened 2026-09-08
feat(runtime): discover pinned Codex in an isolated offline profile

RepairYourTech:
Adds a versioned, bounded runtime inventory and a real read-only Codex App Server discovery probe under #186. Qualification rejects stale or mismatched identities and unknown mandatory facts; observed account/model metadata grants no execution or billing authority.

The installed Codex 0.118.0 successfully completes initialization, account/read without refresh, and model/list in the Linux sandbox with a fresh disposable HOME and denied network. It reports authentication required and six listed models; no coding turn starts. CI repeats this proof using an exact Linux binary artifact with SHA-512 verification.

Validation: workspace tests, strict workspace Clippy, formatting and 13 focused discovery tests pass. Independent cross-review fixed empty-worktree enforcement and propagated cancellation errors. The live handshake corrected the initially assumed user-agent prefix; generated version-specific schemas establish optional-field handling.

See docs/contracts/runtime-discovery.md for evidence and limits. #186 remains open: durable Host integration, configured account instances, authentication and entitlement checks, installation consent and actual coding workflows are pending. This does not complete native or external execution or first-release acceptance.

PR #497, opened 2026-09-09
feat(external): add the sandboxed production transport and real-binary handshake proof

RepairYourTech:
Continuation of the #465/#464 external-loop slice (PR #496 merged the driver). This PR closes the biggest honest non-claim from #496: the production transport.

## What exists now
- **Complete pinned protocol handshake**: the driver's `begin_thread` now performs `initialize` → pinned `symbiote/0.118.0` server-version check (single source of truth: the #483 discovery pin, imported from `symbiote-runtime-discovery`) → `initialized` notification → `thread/start`. A server reporting any other version is refused (`UnsupportedVersion`) before any thread, turn or account interaction. Tests assert the exact sent-method sequence and the refusal.
- **New `process` module — the production transport**: `CodexServerProcess<T: FrameIo>` frames versionless JSONL over any IO. Production IO is the sandboxed `SandboxProcess`; tests use the shared `JsonlTransport` over real subprocesses (scripted frames, no codex binary). Strict frame classification: responses must correlate to the outstanding request id (string-or-number echoed verbatim); notifications and server requests buffer with bounds; unknown envelope fields, null ids, stray responses and error responses are explicit violations (`MalformedFrame`/`UnexpectedResponse`/`RpcFailure`). Per-method denial bodies come from the pinned response schemas (`denied`/`decline`/empty-grant/`success:false`); bodyless methods (`item/tool/requestUserInput`, unknowns) stay unanswered — the transport structurally cannot send a grant.
- **`launch_sandboxed`** composes the Host sandbox (#218) with the pinned binary exactly like the #483 discovery proof: trusted helper, consent whose fingerprint covers the exact codex invocation, disposable HOME, isolated network, reserved worktree (#211), `/usr/bin/codex app-server --listen stdio://`.
- **Observation-loop fixes** (found by the first hung test run, then regression-pinned): the silent-poll budget's reset fired unconditionally (infinite loop on a silent harness) — now it only resets on received frames; server requests surface through the same frame pump as notifications and are drained before the notification is processed.
- **`codex_thread_smoke` example — the real-binary proof**: sandboxed launch, real handshake, real `thread/start`, cancellation. Ran locally against the installed `/usr/bin/codex 0.118.0` inside bubblewrap: `{"handshake":"initialize_initialized_ok","thread_started":true,"turn_started":false,"model_called":false,"credentials_mounted":false,"network":"denied_by_sandbox"}`. CI gains this proof alongside the discovery proof.

## Verification
- 28 crate tests (up from 18): fixtures + full driver-over-real-subprocess integration (handshake → thread → turn → events), EOF/process-exit pin, silent-harness pin, framing/violation matrix, denial-body matrix, id-echo envelope tests.
- Workspace: 345 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Live evidence: local smoke run output above; CI repeats it on disposable runners with the SHA-512-verified pinned binary.

## Honest non-claims
- **No model turn ran** — with authentication `required` and no credentials in the sandbox, a turn would fail anyway; execution proof still requires explicit user authorization for credentials and billing. Surfaced, not attempted.
- No tool-result round trips, no steering/interrupt, no durable resume/reconnect, no Host-side wiring yet (dispatch → driver → RequestCompletion integration is the next slice).
- #464/#465 remain open.

### DIFF
--- a/.github/workflows/rust-contracts.yml
+++ b/.github/workflows/rust-contracts.yml
@@ -58,3 +58,12 @@ jobs:
           cargo run --locked -p symbiote-runtime-discovery --example codex_discover -- \
             "$GITHUB_WORKSPACE/target/debug/symbiote-sandbox-launch" \
             "$proof_root/worktree" "$proof_root/protected"
+      - name: Prove pinned Codex driver handshake and thread start without a model turn
+        if: matrix.toolchain == 'stable'
+        run: |
+          cargo build --locked -p symbiote-external-agent --example codex_thread_smoke
+          smoke_root=$(mktemp -d "$RUNNER_TEMP/smoke.XXXXXXXX")
+          mkdir -m 700 "$smoke_root/worktree" "$smoke_root/protected"
+          cargo run --locked -p symbiote-external-agent --example codex_thread_smoke -- \
+            "$GITHUB_WORKSPACE/target/debug/symbiote-sandbox-launch" \
+            "$smoke_root/worktree" "$smoke_root/protected"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,7 +376,11 @@ dependencies = [
  "serde",
  "serde_json",
  "symbiote-domain",
+ "symbiote-runtime-discovery",
  "symbiote-runtime-sdk",
+ "symbiote-runtime-transport",
+ "symbiote-sandbox",
+ "symbiote-trust",
 ]
 
 [[package]]
--- a/crates/symbiote-external-agent/Cargo.toml
+++ b/crates/symbiote-external-agent/Cargo.toml
@@ -7,6 +7,13 @@ publish = false
 [dependencies]
 symbiote-domain = { path = "../symbiote-domain" }
 symbiote-runtime-sdk = { path = "../symbiote-runtime-sdk" }
+symbiote-runtime-discovery = { path = "../symbiote-runtime-discovery" }
+symbiote-runtime-transport = { path = "../symbiote-runtime-transport" }
+symbiote-sandbox = { path = "../symbiote-sandbox" }
+symbiote-trust = { path = "../symbiote-trust" }
 schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+
+[[example]]
+name = "codex_thread_smoke"
--- a/crates/symbiote-external-agent/examples/codex_thread_smoke.rs
+++ b/crates/symbiote-external-agent/examples/codex_thread_smoke.rs
@@ -1,0 +1,109 @@
+//! Offline compatibility proof with fixture consent, not Host activation
+//! authority: launch the pinned Codex App Server inside the Symbiote sandbox,
+//! complete the driver's real handshake (initialize → pinned version →
+//! initialized → thread/start), and stop. No turn starts, so no model is
+//! called, no credential is read (fresh disposable HOME) and nothing is
+//! spent. Usage:
+//!   codex_thread_smoke <helper> <empty-worktree> <protected-dir>...
+use serde_json::json;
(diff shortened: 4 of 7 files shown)


RepairYourTech:
## Review disposition (independent reviewer, fixes in `2695e2fc` + `be0b9882`)

An independent reviewer agent (not the author) reviewed PR #497 at `a059e790` and, importantly, **re-ran the real-binary smoke proof locally** (sandboxed codex 0.118.0, handshake + thread start, no model turn) to confirm it does what it claims. Verdict: **0 P0 / 2 P1 / 6 P2, mergeable after fixes; the hard invariants (no grant path, no credentials/account operations, evidence-only completion, version pin, exit≠completion, sandbox consent/fingerprint binding) all hold and are test-pinned.** All findings are addressed.

### P1 findings → fixes

1. **Failed `begin_thread` left the session retryable; a retry would record a second `Ready` the SDK tracker rejects** (journal corruption on the plausible production retry loop). Fixed: any failure after the Ready event is terminal — the driver records a `TransportLost` stop + diagnostic, and every entry point then refuses. Regression tests: `failed_handshake_is_terminal_and_retry_never_doubles_ready` (three bad-version servers; asserts exactly one Ready, retry/turn/completion refusals, single Ready, tracker replay) and `malformed_thread_response_is_terminal_too`.
2. **Documented worktree/cwd contract contradicted the sandboxed launcher, and the real-binary proof bypassed the driver.** The sandbox mounts the Host path only at `/workspace`; `begin_thread` now documents and takes the in-sandbox cwd, and the smoke example now drives `ExternalSession::begin_thread` over the production `CodexServerProcess` framing instead of a hand-rolled handshake — re-run locally: `{"handshake":"driver_begin_thread_ok","events":1,"thread_started":true,"turn_started":false,"model_called":false,"credentials_mounted":false,"network":"denied_by_sandbox"}`. (Driving the driver path caught the dispatch fixture's grant requirements, which are now correct in the proof.)

### P2 findings → fixes

- `classify` response branch now rejects out-of-shape known keys (response = id + result xor id + error, nothing else).
- A frame arriving just past the poll deadline no longer hard-fails the transport; buffered server requests surface on the next drain.
- A poll that surfaced server requests resets the silence budget — an escalated harness is alive and conversed with, not silent.
- Refusal-send and absorb failures now mark the run terminal (`TransportLost` + diagnostic), consistent with every other in-turn failure path.
- The leftover debug probe test module was removed.
- Budget constants honest: one per-turn constant `MAX_TURN_NOTIFICATION_FRAMES = 1024` (documented), replacing the unenforced per-call pair; docs corrected (foreign frames are ignored with a diagnostic, not "absorbed"); production must size `TransportLimits` explicitly — the shared transport's default 64 KiB frame cap rejects oversized turns as `FrameTooLarge`, and event-text truncation only applies after framing.

Qodo was retriggered (`/review` at both heads); the bot reports "reviews are paused for this user" and produced no review. CodeRabbit posts summaries only. The completed independent review is the reviewer agent's findings above; the reviewer independently re-ran the local smoke proof and got the same result as the author.

### Verification at the fixed head `be0b9882`

- `cargo test --workspace`: 346 passed, 0 failed; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- One CI iteration failed first (Rust contracts 1.85.0 rejected a collapsed if-with-let-chain — E0658 unstable on the pinned MSRV while local stable accepted it — exactly why fresh CI is the MSRV proof); replaced with the nested form and a named constraint, now green at `be0b9882`: Rust contracts / Bounded Linux proofs / Roadmap integrity all success.
- Offline only: the smoke proof starts no turn; no credential read, no spending.

PR #498, opened 2026-09-09
feat(host): wire started dispatches to both worker loops with evidence-only completion

RepairYourTech:
Closes nothing — an integration slice of #465/#205, which stay open.

## What exists now
- New `symbiote-host::runner`: the composition point where a started dispatch is executed by one of the two worker loops (#465) and the loop's completion report is filed as canonical evidence. This closes the "completion wiring" pending item that `docs/contracts/dispatch-preparation.md` tracked for both runtimes.
- **Runtime selection is the contract's, not the caller's**: `run_native` refuses a dispatch whose profile is not `NativeSymbiote`; `run_external` refuses one that is not `ExternalHarness`. No fallback loop exists — native work never touches the harness, external work never touches the native inference path, and neither loop can be pointed at the other's billing route. The native model envelope is read from the #464 provider registry keyed by the contract's profile model, never from loop input.
- **Preconditions re-checked, not trusted**: the task must be `Running` under this exact dispatch per the journaled store before the loop runs; each loop session re-validates the dispatch contract at run time (expired enforcement windows and foreign identities are refused).
- **Completion is evidence, structurally**: on a clean stop the final agent message is filed through the store's journaled `apply_task` as `Actor::Worker(dispatch)` with `TaskAction::RequestCompletion` — the same domain transition the protocol's `request_task_completion` uses. The task can only reach `CompletionRequested`; the runner has no code path to `Complete` or `BeginVerification`. Host verification and independent review remain the completion gates.
- **Failure files nothing**: halted/failed/interrupted/transport-lost loops leave the task `Running` and return `LoopFailed`; retry policy stays with the Host.
- Tests run against an in-process store with the full composition fixture (project, Team, provider registry entry + model descriptor, workforce binding, classified Objective origin, task, route, preparation, start) and scripted transports for both runtimes: evidence filing on both paths (task reaches `CompletionRequested`, never `Completed`), foreign-runtime refusals leave `Running`, failed-loop no-ops, not-Running refusal before any loop.
- Contract: `docs/contracts/worker-completion-wiring.md`.

## Verification
- Workspace: 351 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Offline only: fixtures prove the wiring; no live API call, no credential read, no spending.

## Honest non-claims
- The transports are deterministic fixtures; live native Responses-API and live Codex model turns each require explicit user authorization for credentials and billing (surfaced, not attempted).
- The runner is a library composition point: the daemon does not yet expose a worker-activation endpoint and nothing schedules runs automatically. Durable session resume, tool execution, and multi-turn tool-result round trips remain pending.
- #464/#465 remain open.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,11 +388,17 @@ name = "symbiote-host"
 version = "0.1.0"
 dependencies = [
  "nix",
+ "rusqlite",
+ "schemars",
  "serde",
  "serde_json",
  "symbiote-domain",
+ "symbiote-external-agent",
  "symbiote-host-inventory",
+ "symbiote-native-agent",
  "symbiote-protocol",
+ "symbiote-runtime-discovery",
+ "symbiote-runtime-sdk",
  "symbiote-store",
  "symbiote-workforce",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,7 +388,6 @@ name = "symbiote-host"
 version = "0.1.0"
 dependencies = [
  "nix",
- "rusqlite",
  "schemars",
  "serde",
  "serde_json",
--- a/crates/symbiote-host/Cargo.toml
+++ b/crates/symbiote-host/Cargo.toml
@@ -11,6 +11,12 @@ symbiote-host-inventory = { path = "../symbiote-host-inventory" }
 symbiote-domain = { path = "../symbiote-domain" }
 symbiote-protocol = { path = "../symbiote-protocol" }
 symbiote-store = { path = "../symbiote-store" }
+symbiote-native-agent = { path = "../symbiote-native-agent" }
+symbiote-external-agent = { path = "../symbiote-external-agent" }
+symbiote-runtime-sdk = { path = "../symbiote-runtime-sdk" }
+symbiote-runtime-discovery = { path = "../symbiote-runtime-discovery" }
+schemars.workspace = true
+rusqlite = { version = "=0.40.2", default-features = false, features = ["bundled"] }
 serde.workspace = true
 serde_json.workspace = true
 nix = { version = "=0.30.1", features = ["socket", "user", "fs"] }
--- a/crates/symbiote-host/Cargo.toml
+++ b/crates/symbiote-host/Cargo.toml
@@ -16,7 +16,6 @@ symbiote-external-agent = { path = "../symbiote-external-agent" }
 symbiote-runtime-sdk = { path = "../symbiote-runtime-sdk" }
 symbiote-runtime-discovery = { path = "../symbiote-runtime-discovery" }
 schemars.workspace = true
-rusqlite = { version = "=0.40.2", default-features = false, features = ["bundled"] }
 serde.workspace = true
 serde_json.workspace = true
 nix = { version = "=0.30.1", features = ["socket", "user", "fs"] }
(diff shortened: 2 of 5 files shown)


RepairYourTech:
## Review disposition (independent reviewer, fixes in `f90fcf3d`)

An independent reviewer agent (not the author) reviewed PR #498 at `521e51f2`, reading the domain transition table, the store's apply_task, and both loop crates to verify the invariants rather than trusting the doc. Verdict: **0 P0 / 3 P1 / 4 P2 — no invariant hole: the domain's transition law holds regardless of runner bugs; nothing can pass `CompletionRequested` through this wiring.** The reviewer also confirmed mutation sensitivity: each of the four hard invariants would fail a specific test if the wiring broke. All findings are addressed in `f90fcf3d`.

### P1 findings → fixes

1. **Cancelled native turns could file completion evidence** — the native loop maps `FinishReason::Cancelled` to the same `HaltReason::Stop` as a normal finish, so the runner's halt-reason gate passed a cancelled turn and would have filed its partial text as worker-complete. Fixed: the runner additionally checks the stream's terminal events (no `CancelAcknowledged`) before filing. Regression test `native_cancelled_turn_files_nothing` drives a `CancelledTransport` whose partial text ("I'll continue implementing...") must never be filed; the task stays `Running`.
2. **Deterministic filing command id made a Host retry a silent no-op** — `worker-completion-{task_id}` could be replayed by the store's idempotency layer, reporting success for work not observed. Fixed: the id now carries the dispatch id and filing time (`worker-completion-{dispatch}-{at}`), so retrying the same observed run replays byte-identically while a new run's filing can never be swallowed. The store error cause is preserved (`RunnerError::Store(cause)`) instead of discarded, making an idempotency collision distinguishable from a state move.
3. **Loop error identity was collapsed into a bare `LoopFailed`** — caller-sequencing bugs (`invalid_input`, `already_complete`) were indistinguishable from retryable `provider_failed`. Fixed: `RunnerError::LoopFailed(reason)` carries the loop's own halt/error name for both runtimes (full name tables for `LoopError`, `DriverError`, `HaltReason`, `StopKind`), and the doc's run-summary overclaim was replaced with the actual contract.

### P2 findings → fixes

- Dead `completion_requires_running` folklore helper removed (the domain transition table is the authority; the runner never needed a shadow copy).
- Unused `rusqlite` host dependency removed (would have pulled the bundled C toolchain into the host build for nothing).
- The fixture's hand-built Role is now pinned by an explicit identity assertion: the store's own `start_prepared_task` re-reads role/binding from its tables and compiles its own dispatch, and the fixture asserts the started dispatch id equals the fixture's — a divergence fails loudly instead of testing against a phantom dispatch.
- New regression tests: `external_completed_turn_without_message_files_nothing` (the `NoReport` path, avoiding an empty report the domain would reject) and `external_interrupted_turn_files_nothing` (the `Interrupted` stop kind).

Qodo was retriggered (`/review` at both heads); the bot reports "reviews are paused for this user" and produced no review. CodeRabbit posts summaries only. The completed independent review is the reviewer agent's findings above.

### Verification at the fixed head `f90fcf3d`

- `cargo test --workspace`: 354 passed, 0 failed; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- CI green at `f90fcf3d`: Rust contracts / Bounded Linux proofs / Roadmap integrity all success.
- Offline only: fixture transports; no live API call, no credential read, no spending.

PR #499, opened 2026-09-09
feat(host): add the worker-activation endpoint with transport-authorized execution

RepairYourTech:
Continuation of the #205/#465 dispatch loop: PR #498 added the runner library composition; this PR exposes it as a real protocol operation. Closes nothing — #205/#465 stay open.

## What exists now
- **Protocol v1.13** (`run_started_dispatch`): activates an already-started dispatch through the Host runner. The loop runtime is selected by the dispatch contract's runtime kind (strict native/external separation preserved); the task prompt is resolved from canonical state — the task origin's classified work-item description — never from caller input. The dispatch binding is re-verified against journaled state before anything runs; a foreign dispatch id is refused with a precondition error before any loop starts. Goldens, version-rejection lists and all checked-in fixtures moved to v1.13 in lockstep.
- **Transport authorization is the boundary**: transports are injected through `WorkerTransports` factories. The production configuration is empty — activation refuses with a typed error ("no worker transport configured for this runtime kind") rather than executing anything, because live native and external execution each require the sandboxed launch path plus explicit operator/user authorization for credentials and billing. There is no fallback loop to fall back to; runner errors map to stable protocol messages that never echo task content.
- **Both loop crates now accept `?Sized` transports** (blanket `&mut T` impls), so the service holds `Box<dyn ...>` factories and dispatches object-safely.
- **Daemon-level test** walks the full protocol surface on a real spawned daemon: register → team (stream-mutation grant patched) → workforce binding (eligible host patched to the live inventory identity) → provider connection → model descriptor → classified objective origin → task → route → preparation (Ready) → start (Running) → activation refuses with the no-transport message, the task stays `Running`, a foreign dispatch id is refused before anything runs, and state replays identically after a daemon crash/restart.

## Verification
- Workspace: 355 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Offline only: no live API call, no credential read, no spending; the no-transport refusal is the tested behavior, not an assumption.

## Honest non-claims
- No live transport is wired into the daemon: a started dispatch cannot yet actually execute a model turn — that requires the sandboxed launch path (#211/#218) plus explicit operator/user authorization for credentials and billing (surfaced, not attempted).
- No scheduling: runs are requested explicitly; nothing queues or activates automatically. No streaming, no tool execution, no durable in-flight session resume.
- #464/#465/#205 remain open.


RepairYourTech:
## Review disposition (independent reviewer, fixes in `86b83445`)

An independent reviewer agent (not the author) reviewed PR #499 at `388cd447`, verifying the invariants against the protocol, service, runner and both loop crates. Verdict: **0 P0 / 1 P1 / 4 P2 — the security invariants all hold: local-owner authority, foreign-dispatch-before-build, typed no-transport refusal with zero content echo, canonical-state prompt provenance, strict production transport emptiness.** All findings are addressed in `86b83445`.

### P1 finding → fix

**Task-state (Running) verification happened after the transport build.** The service arm checked only the dispatch-id binding before invoking a transport factory; the Running state was checked solely inside the runner (`check_runnable`), after the factory had already been invoked — so once an operator provisions a transport, re-activating a task that already filed completion would have launched a transport it could never use. Fixed: both preconditions (dispatch-id binding and Running state) are now checked before any transport build, with a dedicated refusal message ("task is not running under the requested dispatch") that makes the ordering observable. The daemon test pins that exact message on the foreign-dispatch case, which previously asserted only the error code — so a future reordering that built transports first would now fail the test.

### P2 findings → fixes

- `WorkerRun` doc corrected: a halted turn is an `Err` response (the task stays Running; retry policy is the Host's), so `completed: true` in every reachable `Ok` is the journaled-evidence marker, never a verified completion — no phantom `completed: false` state for client authors to build on.
- The serve loop carries a known-limitation note: connections are served synchronously; if an operator ever provisions a transport, turns must move off the connection loop or every connection stalls for the turn's duration (unreachable today — production transports are empty).
- New runner test `transport_factories_build_and_drive_both_runtimes`: the service→factory→boxed-runner chain is now exercised end to end for both runtimes (`with_native`/`with_external` → `native_build`/`external_build` → `run_*_boxed` → task reaches `CompletionRequested`), plus the empty-factory typed `NoTransport` refusal for both kinds.

### Verification at the fixed head `86b83445`

- `cargo test --workspace`: 356 passed, 0 failed; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- CI green at `86b83445`: Rust contracts / Bounded Linux proofs / Roadmap integrity all success.
- Offline only: no live transport is wired into the daemon; the no-transport refusal is the tested behavior, not an assumption. No live API call, no credential read, no spending.

PR #501, opened 2026-09-09
feat(sdk): add the shared client state foundation with journal-cursor resume

RepairYourTech:
First slice of H03 (#182): one implementation of the client-side facts instead of each controller recreating synchronization logic. Closes nothing — #182 stays open.

## What exists now
- New crate `symbiote-client-sdk`. `ClientSession` builds versioned requests pinned to `CURRENT_VERSION`, uses the caller's `command_id` verbatim as the durable idempotency key, mints unique correlations, and refuses oversized requests (64 KiB) client-side after applying the same parse the daemon will.
- **Response classification is exactly four terminal outcomes**: `Ok(typed ResponseBody)`; `Refused(daemon's typed error)`; `Transport` (disposition unknown — never treated as an outcome); `Unparseable` (including a response version that no longer matches — the daemon was replaced under us; re-handshake, never guess). A disconnect is a transport fact, never a task outcome: the SDK never invents worker success.
- **Journal cursor resume**: `Journal` responses advance a per-Project durable cursor over events actually returned; positions survive process restart via `with_positions`; unknown Projects bootstrap at cursor 0.
- **Bounded recovery**: `recover` resends the same command id and byte-identical operation after transport failures (the daemon's journal replays the durable receipt), surfaces refusals immediately, and gives up within the caller's attempt budget.
- **No I/O in the crate**: the transport is an injected `FrameExchange` trait, so the Host's authenticated Unix socket, in-process harnesses, and future remote transports share identical state logic.
- **Deterministic chaos test**: a fixed LCG-scheduled hostile sequence (transport failures, unparseable frames, wrong-version responses, refusals, durable receipts) asserts every terminal classification is honest and durable successes observe the exact journal cursor. Reproducible on every run and platform.
- Contract: `docs/contracts/client-sdk.md`.

## Verification
- Workspace: 364 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Offline only; no I/O in the crate (transport injected).

## Honest non-claims
- No live transport shipped: wiring to the Host socket is the embedding client's job (the CLI currently uses its own direct exchange — a follow-up can migrate it onto this SDK).
- No optimistic-command semantics (revision CAS is surfaced as `StaleRevision` refusals, not speculatively applied), no streaming events, no offline cache, no resync telemetry, and no protocol chaos suite against a live Host. #182 stays open.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,6 +350,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "symbiote-client-sdk"
+version = "0.1.0"
+dependencies = [
+ "serde",
+ "serde_json",
+ "symbiote-domain",
+ "symbiote-protocol",
+]
+
+[[package]]
 name = "symbiote-config"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 resolver = "3"
-members = ["crates/symbiote-worktrees", "crates/symbiote-native-agent", "crates/symbiote-external-agent", "crates/symbiote-domain", "crates/symbiote-config", "crates/symbiote-architecture", "crates/symbiote-store", "crates/symbiote-protocol", "crates/symbiote-host", "crates/symbiote-runtime-sdk", "crates/symbiote-runtime-transport", "crates/symbiote-projection", "crates/symbiote-trust", "crates/symbiote-sandbox", "crates/symbiote-runtime-discovery", "crates/symbiote-host-inventory", "crates/symbiote-workforce"]
+members = ["crates/symbiote-worktrees", "crates/symbiote-native-agent", "crates/symbiote-external-agent", "crates/symbiote-client-sdk", "crates/symbiote-domain", "crates/symbiote-config", "crates/symbiote-architecture", "crates/symbiote-store", "crates/symbiote-protocol", "crates/symbiote-host", "crates/symbiote-runtime-sdk", "crates/symbiote-runtime-transport", "crates/symbiote-projection", "crates/symbiote-trust", "crates/symbiote-sandbox", "crates/symbiote-runtime-discovery", "crates/symbiote-host-inventory", "crates/symbiote-workforce"]
 
 [workspace.package]
 version = "0.1.0"
--- a/crates/symbiote-client-sdk/Cargo.toml
+++ b/crates/symbiote-client-sdk/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "symbiote-client-sdk"
+version.workspace = true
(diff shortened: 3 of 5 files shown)


RepairYourTech:
## Review disposition (independent reviewer, fixes in `dda46eda`)

An independent reviewer agent (not the author) reviewed PR #501 at `0b869c14`, including a step-by-step simulation of the chaos schedule and Host-side parse-parity checks. Verdict: **0 P0 / 3 P1 / 4 P2 — the headline invariant (no fabricated success or cursor; Ok only with a body the daemon actually sent) holds by construction.** All findings are addressed in `dda46eda`.

### P1 findings → fixes

1. **The oversized-request test failed earlier on an invalid operation kind and never reached the size bound** — it would have passed with the size check deleted. Fixed: the test now pads a *valid* `request_task_completion` over 64 KiB and additionally asserts the transport was never touched.
2. **`ClientError::RequestTooLarge` was a dump for four unrelated failure modes with a false doc comment** — command-id charset failures and operation shape/validation failures now have their own variants (`InvalidCommandId`, `InvalidOperation`), each with accurate docs, so a caller branching on "too large" to shrink a payload cannot loop forever. New tests pin each identity, including a semantic-validation failure (`limit: 0`) that is a 40-byte well-formed request.
3. **The chaos test claimed wrong-version frames and full-schedule consumption it did not deliver** (the reviewer simulated the xorshift schedule and showed it died at frame 3). Rewritten: all five frame classes (transport failures, unparseable frames, wrong-version responses, refusals, Journal receipts) with schedule assertions that each class appears at least once; every frame consumed; the frame count must equal the sum of honest classifications; and the durable cursor must equal the *last genuinely observed* Journal page — fabricated successes would fail the assertion.

### P2 findings → fixes

- A `Journal` page is only observed when the caller's project matches the operation's own project — cross-project cursor contamination from a mismatched hand-off is now impossible (pinned by tests).
- The version-mismatch → `Unparseable` mapping is documented on the variant itself: the disposition is genuinely unknown, so id replay stays the honest recovery.
- Contract-doc imprecisions fixed: structural version pinning (no construction-rejection phrasing), the project-hand-off precondition for cursor advancement, and xorshift64 (not LCG) named as the mechanism.
- New pins: a non-Journal `Ok` body and a `Journal` response with `project: None` leave positions untouched.

### Verification at the fixed head `dda46eda`

- `cargo test --workspace`: 365 passed, 0 failed; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- CI green at `dda46eda`: Rust contracts / Bounded Linux proofs / Roadmap integrity all success.
- Offline only: the crate performs no I/O (transport injected).

PR #503, opened 2026-09-09
feat(repo): add dispatch-time worktree provisioning with base-commit validation

RepairYourTech:
Continuation of #211 (reserved worktree locations) + #190 (materialization, PR #502): the composition that turns a started dispatch into a materialized, base-validated worktree. Closes nothing — #211/#190 stay open.

## What exists now
- **New `symbiote-repo::provision`**: the three-stage provisioning a started dispatch runs before any work happens — (1) claim and verify the Change Stream's reserved location (identity re-derived from canonical Project/Root/Stream + policy seed; a mismatch with the stream's recorded worktree id is tamper detection and refuses before any git call), (2) validate the source repository's actual HEAD against the stream's recorded base commit (a moved base means a stale premise: refusal, never materialization against unknown ground), (3) materialize the worktree at the stream's derived branch.
- **Base validation is the security contract**: a moved base refuses with `BaseMoved` *before* materialization — pinned by a test where the source repo advanced past the recorded base and the reserved location is asserted to hold no git content.
- **Materialize redesigned for the real flow**: the work branch is CREATED explicitly at the validated base (`worktree add -b <branch> -- <path> <base-sha>`), resolving the reviewer's DWIM finding structurally — creation is explicit and documented, no remote-tracking ambiguity. `ChangeStream` gains `base()`/`worktree_id()` accessors so the evidence trail uses the recorded identities.
- Errors are stage-honest (`Reservation` / `BaseUnresolved` / `BaseMoved` / `Git` / `NoHostPath`), so retry policy can distinguish a compromised location from a stale premise from a git refusal.
- Tests: three provisioning tests against a real source repository — happy path (materialized at recorded base/branch with content assertions and an honest double-provisioning refusal via the reservation layer's non-empty rule), moved-base refusal, seed-mismatch tamper refusal.
- Contract: `docs/contracts/repository.md` (provisioning stage ordering and base validation documented).

## Verification
- Workspace: 374 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Offline only: local temporary repositories; no remote contact.

## Honest non-claims
- The runner does not yet call `provision` automatically (the daemon's activation endpoint refuses without configured transports); wiring it into `run_external`/`run_native`'s worktree hand-off is the next slice once a source-repo placement source is provisioned by the operator.
- No rebase/revalidation, no synchronization, no integration state, no evidence-confirmed cleanup. #211/#190 stay open.

PR #504, opened 2026-09-09
feat(host): wire worktree provisioning into dispatch activation

RepairYourTech:
Direct continuation of #503 (provisioning composition) — this wires it into the started-dispatch path. Closes nothing — #205/#211 stay open.

## What exists now
- **`RunStartedDispatch` provisions the worktree**: after the precondition checks and BEFORE any transport is built, the service arm runs `runner::provision_worktree`, which resolves everything from trusted store state — the task's Change Stream (worktree id, derived branch, recorded base commit) via a new integrity-checked `task_stream()` reader, the Root record's host-path placement for this Host via a new `root()` reader, and the Host-owned policy seed (validated by symbiote-worktrees' own seed rules) — and runs the #211 composition: reservation verify, base validation against the stream's recorded base, materialization at the derived branch. A moved base or tampered location refuses before any transport factory runs; errors map to typed protocol messages that never echo task content.
- **Configuration boundary**: the production `WorkerTransports` configures no reservation base — activation refuses with "no worktree reservation base configured" before any store read or git call, and the daemon test pins exactly that ordering. Live execution still requires explicit operator/user authorization for credentials and billing (surfaced, not attempted).
- **Tests**: provision_worktree refuses roots without this Host's placement (before any git call); provision_worktree materializes from store records end to end against a real git repository with a placed Root (observed head == base sha, branch state pinned, base content present, honest re-provisioning refusal via the reservation layer's non-empty rule). The runner fixture gains `full_fixture_with_placement` (host_paths placement + base override) so the stream records what the host derivation reproduces. The daemon test's refusal pin moves to the new production ordering.

## Verification
- Workspace: 377 tests pass; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Offline only: real local git repositories; no remote contact, no credentials, no spending.

## Honest non-claims
- Live transports remain unwired: with the production configuration, activation refuses at the provisioning stage before any model turn. Wiring the sandboxed launch into the transport factories is the next slice and requires explicit operator/user authorization for credentials and billing.
- The worktree policy seed is a stable per-stream function of the stream id; the store does not yet derive worktree/branch server-side at task creation (it records caller-supplied values and provision's identity re-check refuses mismatches). No streaming, no tool execution, no resume. #205/#211 stay open.


RepairYourTech:
## Review disposition (independent reviewer, fixes in `0ed15d08`)

An independent reviewer agent (not the author) reviewed PR #504 at `24816734`, verifying the wiring order in the service arm, the store readers' integrity checks, the host identity source, git argv handling, and the NoReservationBase ordering. Verdict: **0 P0 / 2 P1 / 6 P2 — no invariant hole: every failure mode fails closed with typed `failed_precondition` refusals before any transport build, and the ordering contract is real and pinned.** All findings are addressed in `0ed15d08`.

### P1 findings → fixes

1. **The Host policy seed embedded the raw stream id, breaking every legal stream id of 61–128 bytes** (domain ids allow 128; the worktrees seed bound is 64), with a comment claiming a truncation the code never performed. Fixed: the seed is now a 64-hex-char `Fingerprint` digest of the stream id — deterministic per stream, always legal charset and length — with a dedicated `InvalidSeed` variant and protocol message. The false comment replaced with the actual digest rationale.
2. **Provisioning-stage store/seed failures surfaced as "completion filing refused by store"** — a message written for the completion-filing path, misattributing the stage to operators debugging refused activations. Fixed: a distinct `ProvisioningStore` variant carries provisioning-time store failures, and `worker_error` splits the `Store` payload so activation refusals are stage-honest on the wire (seed rejection, root integrity, and general store operations each have their own message).

### P2 findings → fixed or noted

- `WorktreeProvisioned`'s "recorded for the dispatch's evidence trail" claim corrected: recording into the journal is future work — today the caller may log it, nothing journals it yet.
- `Store::root` gains the column-vs-body integrity parity its sibling readers have.
- `assert_task_in_stream` adds reader-side defense-in-depth (the stream's task set must contain the task).
- The redundant `RootId` re-wrap and its dead error path removed; `stream.root_id()` passed directly.
- Docs updated: `host.md` (activation now provisions worktrees before any transport would run; production refusal messages), `protocol.md` (v1.13's `run_started_dispatch` described with the provisioning stage), `repository.md` (the Host wiring, the hex-digest policy seed rationale, and the operational consequence the reviewer flagged: a failed run's task cannot re-activate without an explicit `release()`, reported at the Reservation stage).
- The re-provisioning pin tightened to the exact `Reservation` stage (the old disjunction accepted a wrong-stage `Store` error).
- The overbroad "before any store read" comment corrected (precondition reads necessarily precede it); `workers.git()` made an honest associated-style accessor.

### Verification at the fixed head `0ed15d08`

- `cargo test --workspace`: 377 passed, 0 failed; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- CI green at `0ed15d08`: Rust contracts / Bounded Linux proofs / Roadmap integrity all success.
- Offline only: real local git repositories; no remote contact, no credentials, no spending.

PR #507, opened 2026-09-10
feat(host): wire the sandboxed shell executor into dispatch activation

RepairYourTech:
Closes #218 (slice: the sandbox launch executor composition behind `WorkerTransports`; #466 approval flows and #465 remain open).

## What this does

`run_started_dispatch` can now drive real sandboxed shell tool turns in the reserved worktree. `symbiote-host::shell_executor` composes the production `ShellToolExecutor` over `symbiote-sandbox::launch` — WorktreeWrite profile, the dispatch's reserved worktree, no network, private HOME, literal argv.

**Why the output bridge is a worktree file.** The sandbox child's stdout is the transport's JSONL frame pipe: `launch` consumes the ready marker, and `read_stdout` faults on any non-JSON line, so raw tool output cannot flow through it. The executor wraps the invocation in a FIXED `/usr/bin/sh` script — `p="$1"; shift; exec "$p" "$@" > .symbiote-tool-output 2>&1` — so the model's program and arguments become positional parameters (never quoted into a shell string, no expansion anywhere), and tool stdout+stderr land in a file inside the worktree that the Host reads back bounded (256 KiB) and deletes after the read. The wrapper script is a compile-time constant, which is what makes operator consent authorable: the descriptor's variable part is exactly the model's program and literal arguments.

**Consent is received, never minted.** `WorkerTransports::with_shell_executor` installs a `ShellExecutorFactory`; the service builds it with the run's trusted identities (root/host/project/role/profile/binding access — all from the dispatch contract and the provisioning outcome, never loop or model input). The factory's `ShellConsentAuthority` is consulted PER invocation with the exact wrapped descriptor; the returned consent must fingerprint the exact command — the sandbox independently re-derives and enforces that equality at launch, so a loose consent cannot execute. A refused invocation is `ToolExecError::Refused`, recorded as ToolFailed with the constant phrase `shell tool refused: operator consent does not cover this command` (no command echo, no task content) and fed back to the model while the run continues; a configured factory that fails to build is the typed `RunnerError::ShellExecutorBuild` before any loop runs. With no factory configured, declared shell tools stay propose-only (the #506 behavior, unchanged).

**NativeSession ownership fix.** `with_tool_execution` now takes `Box<dyn ShellToolExecutor>` instead of a leaked `&'static mut`; `execute_shell_tool` takes `&mut dyn`. `run_native_boxed` gains `Option<ToolExecution>`; `run_native` passes `None` (signature unchanged for existing callers).

## Verification

[...]

PR #508, opened 2026-09-10
feat(context): resolve run context and scoped credential leases (#217 slice)

RepairYourTech:
Closes nothing — slice of #217 (context/credential resolution; the full secret-broker scope stays open). Serves #269's #217 dependency and both worker loops' context path.

## What this does

Dispatch activation now resolves the run's full context and credential authorization from trusted store state:

- **Context**: `resolve_context` composes the origin work item's description, requirements, constraints, risks and acceptance — plus the dispatch contract's `ContextPolicy` budget and `AccessSnapshot` — into a `ResolvedContext`; `render_prompt` renders the loop's task prompt. Previously `origin_prompt` carried only the description; the structured guidance existed in the store but no run ever received it. All text bounded per item (8 KiB, visible marker) and per list (64 items / 64 KiB total); structural overflow is the typed `Overbound` refusal, never silent narrowing. Both runtimes receive the same resolved prompt.
- **Credentials**: `CredentialBroker` — the operator's process-local registry. A 5-minute `CredentialLease` is issued ONLY when the dispatch's profile references the credential, the owning project matches the dispatch's project, and the environment label is granted by the binding access. Values are zeroized on drop, never serialized, never in Debug. **Revocation is sticky per reference id** — re-registration is refused; rotation is a NEW reference id (found by my own adversarial test: my first draft let same-project re-registration clear the revoked record, which would resurrect a compromised secret).
- **The gate**: the native arm refuses a credential-referencing dispatch with a typed `NoCredentialBroker` precondition when the operator has not configured a broker — proceeding without a declared credential would be a silent downgrade. External harness runs get NO native leases (harness credentials stay harness-owned; the #217 boundary), enforced by construction (the lease path is native-only).

## Honest non-claims

Lease materialization into loop transports is explicitly pending — the native loop has no env-injection mechanism yet, so the lease is held for the run boundary and dropped (documented in service and contract doc). OS keychain/encrypted-at-rest storage, durable broker persistence, redaction corpus, active-process revocation reporting remain with #217. No live model turn exists; fixture values only; **nothing was spent**.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings` clean; `cargo test --workspace --locked` — 69 suites, 0 failures.
- Context crate: resolution bounds/truncation on char boundaries/typed refusals (missing/empty origin, over-bound lists at the REAL 9-item overflow boundary — the first test draft had the boundary wrong and the code proved it), lease issue/expiry-window (including backdated materialization refusal)/revocation stickiness/cross-project/unreferenced/ungranted distinct refusals, Debug value-leak assertions.
- Host: store-backed resolution end to end through the full fixture into the loop's received prompt (requirements/acceptance sections asserted in the transport), `NoCredentialBroker` and `UnknownReference` typed mappings, task stays Running on refusals.

## Review record

An independent reviewer round will be posted as a PR comment (per the audit-trail practice from #507), and merge happens only on its clean verdict.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,6 +428,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "symbiote-context"
+version = "0.1.0"
+dependencies = [
+ "schemars",
+ "serde",
+ "serde_json",
+ "symbiote-domain",
+ "zeroize",
+]
+
+[[package]]
 name = "symbiote-domain"
 version = "0.1.0"
 dependencies = [
@@ -459,6 +470,7 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
+ "symbiote-context",
  "symbiote-domain",
  "symbiote-external-agent",
  "symbiote-host-inventory",
@@ -721,6 +733,12 @@ checksum = "df79d97927682d2fd8adb29682d1140b343be4ac0f08fd68b7765d9c059d3945"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "zeroize"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e13c156562582aa81c60cb29407084cdb54c4164760106ab78e6c5b0858cf64e"
 
 [[package]]
 name = "zmij"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 resolver = "3"
-members = ["crates/symbiote-worktrees", "crates/symbiote-native-agent", "crates/symbiote-external-agent", "crates/symbiote-client-sdk", "crates/symbiote-repo", "crates/symbiote-domain", "crates/symbiote-config", "crates/symbiote-architecture", "crates/symbiote-store", "crates/symbiote-protocol", "crates/symbiote-host", "crates/symbiote-runtime-sdk", "crates/symbiote-runtime-transport", "crates/symbiote-projection", "crates/symbiote-trust", "crates/symbiote-sandbox", "crates/symbiote-runtime-discovery", "crates/symbiote-host-inventory", "crates/symbiote-workforce"]
+members = ["crates/symbiote-context", "crates/symbiote-worktrees", "crates/symbiote-native-agent", "crates/symbiote-external-agent", "crates/symbiote-client-sdk", "crates/symbiote-repo", "crates/symbiote-domain", "crates/symbiote-config", "crates/symbiote-architecture", "crates/symbiote-store", "crates/symbiote-protocol", "crates/symbiote-host", "crates/symbiote-runtime-sdk", "crates/symbiote-runtime-transport", "crates/symbiote-projection", "crates/symbiote-trust", "crates/symbiote-sandbox", "crates/symbiote-runtime-discovery", "crates/symbiote-host-inventory", "crates/symbiote-workforce"]
 
 [workspace.package]
 version = "0.1.0"
@@ -12,6 +12,7 @@ publish = false
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 schemars = { version = "1", features = ["derive"] }
+zeroize = "1"
 
 [workspace.lints.rust]
 unsafe_code = "forbid"
--- a/crates/symbiote-context/Cargo.toml
+++ b/crates/symbiote-context/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "symbiote-context"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+publish.workspace = true
+
+[dependencies]
+symbiote-domain = { path = "../symbiote-domain" }
+schemars.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+zeroize.workspace = true
+
+[lints]
+workspace = true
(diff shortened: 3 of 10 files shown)


RepairYourTech:
## Independent review record — round 1

Review method (same as #507): a separate reviewer agent reads the full diff and empirically probes claims (scratch probes compiled against the worktree crates, gauntlet re-run). Findings and disposition (fix commit 0c5eef0):

| # | Severity | Finding (reviewer's evidence) | Disposition |
|---|----------|-------------------------------|-------------|
| 1 | **P0** | The daemon built "environment grants" by Debug-formatting `Permission` enum values (`format!("{g:?}")`) and matching them against secret environment labels. Probed: correctly registered secret (`OPENAI_API_KEY`) → always `EnvironmentDenied`; a secret labeled `ReadRoot` → mis-granted. The slice's primary path could never succeed and the only "pass" was a mis-grant. | **Fixed.** Replaced with the domain's real typed authority: the binding access must grant `Permission::UseCredential` (the dispatch compiler already forces `Control::Credentials` host enforcement when that permission is granted). Environment labels are materialization names only. Fixture extended with `UseCredential` + `Control::Credentials`; new end-to-end seam test drives the real contract scope through `WorkerTransports`. |
| 2 | P1 | Rotation/revocation dropped the old value only at a later map-insert drop — the old value stays readable in freed-but-uncleared heap until then. | **Fixed.** `register`/`revoke` remove-then-drop the previous record immediately (zeroize at replacement); documented the caller's only-owning-`Vec` obligation. |
| 3 | P1 | Four overclaiming comment sites: "materializes into the loop's prompt envelope boundary", "materializes them only inside the sandbox launch composition", "materialized names are cleaned on drop", "revocation makes every materialization fail closed" — none exist (probe: a lease issued before revocation still materializes until expiry). | **Fixed.** All four rewritten to the honest state: lease issued, scope-checked, dropped unmaterialized; injection and outstanding-lease invalidation explicitly pending #217 (disclosed in crate doc + contract doc). |
| 4 | P2 | Time-window and label-validation failures surfaced as `EnvironmentDenied` — destroying the distinct-refusal contract. | **Fixed.** New `Expired`/`NotYetValid`/`InvalidEnvironmentLabel`/`UseCredentialNotGranted` identities, asserted by test. |
| 5 | P2 | Dead code: `broker_configured()` (no callers), `sensitive` field + `is_sensitive` (never read), revoked-record owner (never read), `NotReferencedByProfile` unreachable in the daemon path. | **Fixed** (all but the last removed/simplified; `NotReferencedByProfile` is a real seam check kept — the daemon passes the profile's own reference, so it is a defense-in-depth arm). |
| 6 | P2 | No end-to-end test drove a lease through the real call path with a broker configured (why finding 1 survived). | **Fixed.** `lease_resolution_runs_through_the_real_contract_scope`: real fixture dispatch contract → passing lease (scope asserted verbatim, materialize value checked), `NoCredentialBroker`, `UnknownReference`, `cross_project_denied`, sticky `revoked`. |

### Verified clean by the reviewer (with probes)

All five `LeaseScope` fields come from trusted state (no request input reaches the broker); cross-project registration refused; revoked references cannot re-register under ANY project; `SecretBytes` zeroizes on drop with no Clone/serde and redacted Debug (no plaintext String outlives the lease — `materialize` has no caller); lease window arithmetic saturating and backdating refused; bounds exact (9×8192 > 65536 > 8×8192, char-boundary truncation, whitespace-only description refused); service refusal ordering intact (reservation-base first); both runtime arms share one prompt; gauntlet green; context tests 5× no flakes.

### Round 2

The reviewer is re-verifying every fix empirically against 0c5eef0; merge only on a clean verdict.

RepairYourTech:
## Round 2 — verdict: fixed-with-evidence, merge recommended

The reviewer re-verified every round-1 fix empirically (own probe crate compiled against the worktree, full gauntlet re-run):

| Round-1 finding | Round-2 evidence |
|---|---|
| P0 grants mapping | Scratch probe: secret registered as `OPENAI_API_KEY` **leases** through the real fixture contract and materializes the correct value with `use_credential_granted=true`; the identical resolution with the grant absent → `UseCredentialNotGranted`; arbitrary labels also resolve (labels are names, not grants). Grep: zero occurrences of the old mapping anywhere. |
| P1 scrub | Probe confirms the restore sequence (A → B denial → A with new value → Ok, resolution materializes the NEW value); no move/drop hazards; clippy force-rechecked clean. |
| P1 overclaims | All four sites rewritten; grep for the old phrases: zero hits. Crate doc discloses outstanding-lease behavior; service says "DROPPED UNMATERIALIZED". |
| P2 identities | `Expired`/`NotYetValid`/`InvalidEnvironmentLabel`/`UseCredentialNotGranted` each asserted by test; `broker_error_name` maps all 8 variants with no wildcard arm (exhaustiveness compiler-enforced). |
| P2 e2e provenance | All scope fields from the real dispatch contract/store; only `host_id` synthetic (the broker never reads it — does not weaken the claim). |

**Fixture-widening audit (clean):** `UseCredential` in the binding access *requires* `Control::Credentials` host enforcement at contract compile (domain law, `dispatch.rs`), so the fixture's `Control::Credentials` claims are required coherence, not gratuitous widening; the escalation test still proves inserting `UseCredential` beyond the ceiling is `PrivilegeEscalation`.

**New findings in 0c5eef0 (both fixed pre-merge in 26005595):**
- P2: a refused cross-project registration had scrubbed the owner's record to an empty shell — probed fail-open window where owner resolutions materialized an empty credential. Now the refusal peeks before removing: the owner's record (value included) is untouched; same-owner replacement still scrubs immediately. Crate test added.
- Cosmetic: stale `Revoked` docstring said same-project re-registration was allowed — now states the sticky rule.

**Gauntlet (reviewer-run + post-fix):** fmt clean; clippy `-D warnings` clean; **406 passed / 0 failed** across 69 suites; context tests 5× no flakes. CI green on 7b05ca3c and 0c5eef0.

Merging.

Issue #54, opened 2026-09-03
[HOST-09][Wave 2] Build the `symbiote` headless CLI for Host, Fabric, Project, task, and recovery administration


RepairYourTech:
## Accepted technology direction — 2026-09-08 takeover

Implement the administrative CLI in Rust against the shared Host contracts. Preserve administrative scope separately from native agent CLI #467 while sharing the native implementation where applicable. See #170 and #38.

**Canonical implementation issue.** Parent epic: #157 · Wave: 2 · Audit: #443.

## Accepted v2.4 requirements and corrections

The following incorporates the latest accepted decisions and controls interpretation of retained baseline wording.

### Native CLI scope separation
- Preserve Host/Fabric/Project/task/recovery administration, machine-readable responses and remote attachment here.
- #467 owns the interactive/headless Symbiote CLI coding-agent experience on the shared runtime; do not implement a second model loop in this administrative issue.
- [ ] Administration can inspect and stop native/external goals/children with scoped authorization even if the desktop is closed.

## Outcome
Provide a scriptable, human-readable client for headless Hosts and operators that uses the same authenticated protocol as every other controller.

## Required scope
- Implement commands for Host status, logs, start/stop/drain, pairing, device revocation, Fabric membership, Project open/list, task status/control, service placement, terminals, diagnostics, and recovery.
- Support local IPC and remote endpoint selection without bypassing authorization.
- Provide stable JSON/JSONL output and exit codes for automation plus concise human output.
- Implement interactive confirmation only for human mode; noninteractive mode must require explicit flags/policies.
- Support shell completion, documented environment variables, config profiles, and redaction.
- Use canonical technical terminology in CLI contracts while allowing optional aliases in help/display.

## Non-negotiable invariants
- The CLI may not become a privileged backdoor around normal RPC authorization and audit.

## Acceptance criteria
- [ ] The CLI can control a headless Fedora Host from another paired machine.
- [ ] JSON output is versioned and passes schema fixtures.
- [ ] Dangerous operations fail without explicit noninteractive authorization.
- [ ] The CLI cannot access operations outside its device scopes.
- [ ] Every mutating command returns an idempotency/result identifier.
- [ ] CLI and desktop produce equivalent task state through the same Host protocol.

## Verification
Test local and remote modes, denied scopes, noninteractive dangerous operations, schema stability, reconnect, and parity against desktop Host calls.

## Audit integration and completion contract
[...]

PR #509, opened 2026-09-10
Desktop workflow shell: SDK-driven demo sequence with authenticated root placement (v1.14)

RepairYourTech:
Closes toward #54 (desktop workflow shell on the critical path to the first-release demonstration; also advances the authenticated-assignment portion of #269's dependencies and #205's dispatch loop). Based on refreshed main b30806ff.

## What this does

**The first-release demo path now runs end to end through the protocol.** A new `symbiote-workflow` crate drives the daemon over `symbiote-client-sdk` with a same-UID socket transport: register project → observe repository placement → team/binding/provider/model → objective-owned task → route → prepare → start → `run_started_dispatch` with real worktree provisioning and the sandboxed shell executor → completion evidence and worktree diff. The e2e test SIGKILLs the daemon between the start and run halves and resumes on the same state directory: the project, task, and started dispatch survive in the journal, the driver's journal position resumes, and the run then produces real diff evidence (`produced.txt` in the reserved worktree) while completion stays at `CompletionRequested` — evidence, never verified completion.

**Protocol v1.13 → v1.14: `observe_root_placement`.** The diagnosed gap this slice fills: the wire had no way to record a Root's host placement — `ProjectDraft::validate()` deliberately refuses non-empty `host_paths` ("an observed placement, not identity"), so only Rust fixtures could ever open a repository. The new mutation operation:

- records a placement only after the daemon verifies the claimed `host_id` equals its own inventory identity, canonicalizes the path, and observes it as a real git work tree (`symbiote-repo::observe_head`);
- advances the Root revision under exact CAS (`expected_revision` → stale revisions refuse with `StaleRevision`);
- journals the resulting Root; the replay audit derives the idempotency bytes from the event, validates the revision transition and the recorded host/path entry, and the final root-state cross-check now replays to the latest observed placement, not registration time;
- keeps registration refusal intact: clients still cannot declare placements, only Host-verified observations.

**Honest enforcement-claims fix (found by the e2e):** `host_record` claimed only Filesystem/Process/Cancellation/CompletionAuthority, yet the domain dispatch compiler requires a `Credentials` control claim when a binding grants `UseCredential` (domain law from #217). A host-with-broker now claims `Credentials`; a Host without a provisioned broker never claims it, so a credential-using dispatch refuses at start instead of running unisolated.

**Operator provisioning composition (`--operator-config`):** reservation base, the explicitly labeled fixture model transport (proposing one sandboxed shell tool), the credential-broker registration, and the shell program allowlist — mapped onto the `WorkerTransports` seams by `assemble_operator_transports`.

## Verification

- Full gauntlet locally: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked` — all green (store 52, host daemon 19, protocol 19, workflow e2e 1, plus all prior suites).
- New coverage: store test for CAS/replay/idempotency-conflict/audit-reopen of placements; protocol tests for path-shape refusal, Register-grant authority, and journal lineage (revision transition + recorded-entry tamper); e2e restart/resume with real sandboxed tool execution.
- Version-bump lockstep: all 24 fixture envelopes, negotiation rejected-versions list, golden captured request/response, daemon test envelope, `docs/contracts/protocol.md`.

## Deliberately not here

- Live model turns for both runtimes (native Responses API, Codex App Server) remain gated on explicit user authorization for credentials and billing — the fixture transport is labeled as such in code and docs. No spending, credential changes, or paid fallback.
- The desktop GUI shell itself (Tauri/workbench) is out of scope; this is the control-plane driver the GUI will call.
- Project-isolation switching (#…) and multi-root placement flows remain open.


RepairYourTech:
## Independent review record — round 1 (head 5fc2f6cd)

Reviewer: independent local reviewer agent (full adversarial scope: security model, store/audit tamper scenarios, protocol mirror, version lockstep, honesty of claims). Reviewer ran the scoped gauntlet at head: `cargo fmt --all --check` PASS; `cargo test -p symbiote-store root_placement` PASS; `cargo test -p symbiote-protocol` PASS (19); `cargo test -p symbiote-workflow --test end_to_end` PASS (run twice, genuine SIGKILL/restart with real `produced.txt` asserted from disk); scoped `cargo clippy ... -D warnings` PASS (incl. symbiote-host beyond scope).

**Verdict: APPROVE. No P0, no P1.**

Hunts that came up empty (evidence cited by reviewer): host-id spoofing (authorize checks Register, service re-checks `inventory.host_id()` before any filesystem/store action), credential-claim gating (only claim path is `has_credential_broker()`, `Dispatch::compile` requires `Credentials` for `UseCredential`), fixture-transport leakage into live-call paths, leftover v1.13 artifacts (only the negotiation test's rejected-versions list, intentional), and the store/protocol `RootPlacementObserved` mirrors are field-identical.

### Findings and dispositions

| F-1 | P2 | New audit arm never validated the journal row's `revision` column against the payload (all sibling arms do). | **Fixed in be59d89b** — arm rejects `revision != root.revision.0`. Additionally, the follow-up tamper test revealed the journal's `journal_no_update` trigger already refuses any row rewrite at the schema level (`RAISE(ABORT, 'append-only journal')`), so the tamper path the reviewer described was doubly blocked; the test now pins that trigger behavior and the audit check stands as the belt to that braces. |
| F-2 | P3 | Service arm runs canonicalize+observe_head before the store replay check; exact retry of a committed command whose repo path was deleted returns NotFound instead of the journaled receipt. | **Accepted with comment** — the common crash-retry case (path present) replays byte-exactly; reordering would add a store round-trip to the refusal path. Revisit if a retry-miss ever surfaces in practice. |
| F-3 | P3 | `assemble_operator_transports(config, user_id)` discarded `user_id`. | **Fixed in be59d89b** — parameter dropped; caller simplified. |
| F-4 | P3 | `WorkflowError::Socket` overloaded for local git/derivation failures. | **Follow-up** — dedicated `LocalObservation` variant when the desktop UI consumes the driver. |
| F-5 | P3 | Driver-side cursor restore (`with_positions`) untested; e2e proves daemon-side restart/resume with in-memory driver continuity. | **Follow-up** — test reconstructing the driver from serialized positions across the restart. |
| F-6 | P3 | socket.rs maps its own pre-send validation failures to `TransportFailure` (unreachable via Driver::call; build_request enforces the same bound). | **Follow-up** — distinguishable FrameExchange error or drop the redundant check. |
| F-7 | P3 | 30 s client socket deadline vs. ≤120 s tool runs under the single-connection daemon; recover replay heals, bounds otherwise consistent. | **Accepted** — documented daemon limitation; revisit with concurrent serving. |

PR #510, opened 2026-09-10
Least-privilege dispatch assignment: task-scoped effective access and binding enforcement floors (#269)

RepairYourTech:
First core of #269 (N09, least-privilege Role/tool/secret/permission assignment), landed at the single choke point every dispatch passes through: `Dispatch::compile`. Its #217/#203 prerequisites have cores; this makes assignment least-privilege rather than scope-inherited. Based on refreshed main 4e42da09.

## What this does

**1. Per-dispatch access narrowing (least privilege recomputed per Dispatch).** The `WorkforceRuntimeContract` now carries an immutable `effective_access` snapshot: the binding's authorized access narrowed to exactly the Task's Root. The binding snapshot itself is retained as authorization provenance, but what a dispatch OPERATES under is task-scoped. The consumers that actually enforce scope now read the narrowed snapshot:
- the sandbox shell executor's consent snapshot (so tool execution is consented against the task's root scope, not the Role's every root);
- context resolution's access source;
- the runner's executor-input composition.

**2. Binding enforcement floors are binding at assignment.** A workforce binding declares `minimum_enforcement: {Control → EnforcementStrength}` (until now consulted only in read-only readiness). `Dispatch::compile` now takes the durable binding's policy floor and refuses the start with `UnsupportedControl` when the Host's enforcement claim is weaker than the declared minimum (same strength ordering as the runtime SDK's minimum-enforcement check, mirrored against claims). Consequences that matter:
- a fallback/reassignment recompiles, so a runtime that cannot meet the policy is **rejected, never inherited**;
- a policy raise between prepare and start is honored at start (start recompiles from the durable binding) — covered by a new store test;
- floors on controls the dispatch does not require are vacuous (Process is only required when `ExecuteProcess` is granted, Network when `Network`, Credentials when `UseCredential` per #217's domain law).

**3. Journal compatibility.** Contracts journaled before narrowing deserialize without the field (serde default, skipped in serialization when absent) and fall back to their compiled binding snapshot — replay of old journals is preserved, verified by round-trip and reopen tests.

## Honest scope against #269's full acceptance

This is the filesystem-scope + enforcement-strength CORE, not the whole issue. Still open under #269: skill/MCP/secret-scope trust metadata and dependency declaration, explicit elevation leases with reason/approver/auto-revocation, production-access as an independent boundary, mid-run policy-change semantics beyond start-time recompile (restart/rebind/deferred), per-denial enforcement-layer reporting, and the two-harness equivalent-permission demonstration.

## Verification

- New tests: domain — narrowing (second binding root excluded from effective access, retained in binding snapshot), floor satisfied/refused/vacuous; store — prepare→raise-floor→start refuses `Domain(UnsupportedControl)`; old-contract-body deserialization fallback.
- The existing store start test now exercises the production floor path end to end (fixture declares HostEnforced floors; start compiles with them).
- Full local gauntlet green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked`, plus the `symbiote-workflow` e2e (SIGKILL/restart demo) passing under narrowed access.
- Still gated on explicit user authorization for credentials and billing: live model-turn verification for BOTH runtimes (native Responses API, Codex App Server). Nothing spent; no fallback to native billing for external execution.


RepairYourTech:
## Independent review record — round 1 (head 00ec9a35, fixes in f09d79df)

Reviewer: independent local reviewer agent (adversarial scope: narrowing enforcement tracing, floor-semantics faithfulness vs the SDK, serialization/replay compatibility, stale-binding-snapshot paths, least-privilege honesty). Reviewer ran the full workspace gauntlet at head: fmt clean, clippy `-D warnings` clean, `cargo test --workspace --locked` exit 0, e2e restart/resume PASS, protocol/fixtures diff empty (no version bump needed — no new ops).

**Verdict: APPROVE. No P0, no P1.**

Key reviewer confirmations: every current enforcement consumer reads `effective_access()` (sandbox consent snapshot, context resolution); the floor check covers required_controls ∪ mandatory ∪ permission-derived controls and is correctly vacuous for non-required ones; `start_prepared_task` reads the CURRENT durable binding (no stale-snapshot path — preparations embed no binding snapshot); the strength ordering and journal compat were verified against the SDK and audit code.

### Findings and dispositions

| F-1 | P2 | `validate_at` did not re-validate `effective_access` — the core narrowing invariant was a convention on the replay path, not a checked invariant. | **Fixed in f09d79df** — `validate_at` now requires effective access to be exactly the Task's Root under the binding's own authorization (project/policy_revision/grants equality, roots ⊇ {root_id} ∩ equal to {root_id}), refusing reconstructed contracts that widen it (`LineageMismatch`). |
| F-2 | P3 | `strength_meets` merged the SDK's incomparable `ExternallyObserved`/`Emulated` minimum arms (looser mirror; unreachable divergence; comment misdescribed it). | **Fixed in f09d79df** — arms split to mirror the SDK arm-for-arm; comment states the reachable-claim argument. |
| F-3 | P3 | `use_credential_granted` still read `binding().access.grants` (equivalent today; future grants-narrowing would silently diverge). | **Fixed in f09d79df** — switched to `effective_access().grants`. |
| F-4 | P3 | Doc did not state that grants are not narrowed, nor the two upgrade semantics (floor activation can newly refuse starts for existing bindings; pre-narrowing journals replay under full binding scope). | **Fixed in f09d79df** — dispatch-preparation.md now covers all three. |
| F-5 | P3 | `skip_serializing_if` means an old binary refuses a new journal (`deny_unknown_fields`). | **Accepted** — the documented no-downgrade stance applies to every journaled field; single-binary local daemon. |

PR #511, opened 2026-09-10
Driver restart resume: serialized journal-position restore test and honest error taxonomy (#54)

RepairYourTech:
Closes the three #509 review follow-ups in the `symbiote-workflow`/`symbiote-client-sdk` crates (F-5 driver-side restore coverage, F-4 error taxonomy, F-6 socket pre-send error identity). Based on refreshed main 00ba543e.

## What this does

**1. The restore path is now proven, not just provided (F-5).** A new e2e test mirrors what a real desktop restart does: generation-1 driver runs the start half, follows the evidence trail, and **persists its journal positions to disk as JSON**; the daemon is SIGKILLed; generation 2 constructs a **fresh driver sharing no memory with the dead one**, restores the serialized positions, and must resume **exactly at the recorded cursor** (asserted before any call), then completes the workflow against the restarted daemon — the run's own events (filed after the restored position) carry the completion evidence, the report, and the worktree diff. The daemon-side SIGKILL/restart proof from #509 plus this driver-side restore together cover both halves of "restart the desktop and resume without losing canonical state".

Supporting API: `ClientSession::positions()` (the persist-side accessor paired with `with_positions`), `JournalPosition` gains a JSON form, and `DemoWorkflow`/`Driver` expose both.

**2. Honest error taxonomy (F-4).** `WorkflowError::LocalObservation` now carries local git read/worktree-derivation/status failures; `Socket` remains strictly a wire-transport/local-framing failure. A desktop UI branching on the variant no longer misreads a filesystem failure as a daemon problem.

**3. No locally-invented unknown dispositions (F-6).** The socket launcher drops its redundant pre-send frame check: the SDK's `build_request` already bounds requests, and the daemon's `parse_request` is the enforcement point — an invalid or oversized frame comes back as a **typed refusal (known disposition)** instead of a locally synthesized transport failure that would trigger pointless recover retries.

## Verification

- New test `driver_restart_restores_serialized_journal_positions_and_resumes` passes alongside the original demo e2e (both SIGKILL-based).
- Full local gauntlet green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked`.
- The shared demo environment was factored into one helper; each test gets a private scratch namespace.

## Standing gate (unchanged)

### DIFF
--- a/crates/symbiote-client-sdk/src/lib.rs
+++ b/crates/symbiote-client-sdk/src/lib.rs
@@ -24,7 +24,8 @@ use symbiote_protocol::{
 /// re-read from `cursor` after reconnect; pages advance it only over events
 /// actually returned. Cursor 0 means "bootstrap from the beginning" — the
 /// caller decides whether a full bootstrap or a tail read is appropriate.
-#[derive(Clone, Debug, PartialEq, Eq)]
+/// The JSON form is what a desktop shell persists across its own restart.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct JournalPosition {
     pub project: ProjectId,
     pub cursor: JournalCursor,
@@ -108,6 +109,12 @@ impl ClientSession {
             .find(|p| &p.project == project)
             .map(|p| p.cursor)
             .unwrap_or(JournalCursor(0))
+    }
+
+    /// The positions to persist across a process restart; restore them
+    /// with [`Self::with_positions`].
+    pub fn positions(&self) -> &[JournalPosition] {
+        &self.positions
     }
 
     fn observe_page(&mut self, page: &Value, project: &ProjectId) {
--- a/crates/symbiote-workflow/src/demo.rs
+++ b/crates/symbiote-workflow/src/demo.rs
@@ -74,6 +74,12 @@ impl DemoWorkflow {
         self.driver.journal_position(project)
     }
 
+    /// The positions to persist across a desktop restart; restore with
+    /// [`Self::with_positions`].
+    pub fn positions(&self) -> Vec<symbiote_client_sdk::JournalPosition> {
+        self.driver.positions()
+    }
+
     fn register_operation() -> serde_json::Value {
         serde_json::from_str::<serde_json::Value>(include_str!(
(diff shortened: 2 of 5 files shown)


RepairYourTech:
## Independent review record — round 1 (head 6179ce44, fixes in 01a58511)

Reviewer: independent local reviewer agent. Reviewer ran the full gauntlet at head (fmt clean, clippy `-D warnings` clean, `cargo test --workspace --locked` green ×3, e2e solo + parallel, SDK tests 8/8) and verified the restore semantics deeply (append-only journal means no compaction path; a restored cursor beyond head yields a typed `InvalidPage` refusal, never a skip; page advance only over returned events; the gen2 read provably starts at `cursor_at_kill`).

**Verdict: REQUEST_CHANGES — accepted and fixed.** No P0/P1. The reviewer caught that the PR's error-taxonomy claim was half-true: the `observe_worktree_evidence` mappings had silently not applied (only the `start_demo` duplicates converted).

| F-1 | P2 | `observe_worktree_evidence` still mapped policy_seed/Derive/observe_status to `Socket` — PR claim #2 was false for those three lines. | **Fixed in 01a58511** — all three now map to `LocalObservation`; zero `WorkflowError::Socket` production mappings remain, so the PR description's taxonomy claim is now accurate. |
| F-2 | P3 | The removal comment misattributed the enforcement point: oversized frames are refused by the daemon's transport frame reader (`Refused(InvalidRequest)`), never reaching `parse_request`'s `RequestTooLarge`. | **Fixed in 01a58511** — comment now names the frame reader and the actual error identity. |
| F-3 | P3 | `exchange` now accepts arbitrary buffers with no documented precondition (EPIPE ambiguity for a hypothetical oversized direct caller; unreachable through the SDK). | **Fixed in 01a58511** — precondition documented on `exchange` (bounded by `MAX_REQUEST_BYTES`, newline-free, guaranteed by `build_request`). |

Reviewer-verified-clean items: journal append-only restore semantics, JSON round-trip shape (plain-string ProjectId, transparent cursor), taxonomy blast radius (no exhaustive matches outside the crate), test independence/parallelism and drop-order cleanup, no dead code from the removed check.

PR #512, opened 2026-09-10
Explicit elevation leases: bounded, attributable capability beyond binding access (v1.15) (#269)

RepairYourTech:
Second #269 core: **explicit elevation with reason, duration/lease, approving authority, and automatic revocation** — the mechanism a Run's mid-run "I need X" becomes a bounded, auditable grant instead of silent inheritance. Based on refreshed main 718cac0f.

## What this does

**Protocol v1.14 → v1.15: `decide_elevation` / `revoke_elevation`.** A decided `ElevationLease`:
- licenses exactly ONE permission beyond a dispatch's binding access — deciding a permission the dispatch **already grants** refuses (`InvalidElevation`: elevation, not re-grant);
- never exceeds the **Team access ceiling** (ceiling refusals map to `failed_precondition`; the ceiling is law even for the owner);
- attributes to exactly one **Running** dispatch (task + dispatch verified against journaled state — every active lease is attributable);
- carries the worker's recorded **reason** and the approving authority;
- is hard-bounded to a ≤15-minute window — **automatic revocation is the expiry check at every use**, no sweeper;
- is **sticky-revocable** (a revocation can never be undone); denials are journaled evidence that license nothing but record who saw and refused the ask.

**Store (schema v11).** `elevations` table + journal events (`elevation_decided` / `elevation_revoked`) with full replay-audit lineage: the decision arm checks attribution against the tasks map at that journal position; the revocation arm re-derives the revocation from the decided lease; the final cross-check replays elevation state from the journal. `active_elevation(dispatch, permission, at)` is the enforcement read.

**Broker integration (one enforcement point wired).** The native run path's credential-lease authority is now "binding grant OR an active elevation lease for THIS dispatch": the owner can license a mid-run `UseCredential` the binding did not grant, without mutating the binding or the Team. The broker itself stays dumb — the Host resolves authority from durable state.

**Authorization.** Deciding/revoking is the authenticated same-UID owner's authority (like resource consent); the store enforces the ceiling/attribution/not-already-granted rules regardless of who asks.

## Honest scope

The worker→Host request channel (a run proposing its own elevation through the protocol) is future scope — this slice makes the DECISION side durable and enforced; the desktop collects the ask and decides. Skill/MCP/secret-scope trust metadata, production-access boundaries, and per-denial enforcement-layer reporting remain open on #269.

## Verification

- New tests: domain (window bounds, expiry-as-auto-revocation, denial licensing nothing, sticky revocation, revocation-before-decision malformed); store lifecycle (route → prepare → start → ceiling raise → approve → active → expired → revoked → sticky-revoke refusal → re-grant refusal → beyond-ceiling refusal → replay → reopen audit); migration drop-lists and future-version probe updated (v11).
- v1.15 lockstep: all 24 fixture envelopes (both spacing variants), negotiation rejected-versions list, golden captured request/response, daemon test envelope, `docs/contracts/protocol.md`.
- Full local gauntlet green (fmt, clippy `-D warnings`, `cargo test --workspace --locked`, both e2e SIGKILL tests).

## Standing gate (unchanged)


RepairYourTech:
## Independent review record — round 1 (head 55af50b4, fixes in the follow-up commit)

Reviewer: independent local reviewer agent. Verdict **COMMENT — no P0, no P1**. Reviewer's adversarial tracing verified clean: ceiling containment (with ceiling changes gated by ManageTeam authority), Running-dispatch attribution at both write and audit time, no cross-project lease-id reuse, dispatch+permission scoping (a lease for dispatch A licenses nothing for B), backdating bounded (a fully backdated window is dead; coverage can never exceed the 15-minute span), malformed stored leases fail closed, and the run-path widening reaches ONLY the broker. Full gauntlet green at head (fmt, clippy `-D warnings`, workspace tests incl. both e2e SIGKILL suites).

### Findings and dispositions

| F-1 | P2 | Elevated runs' contracts carry no `Control::Credentials` claim (the binding did not grant `UseCredential` at compile time), while the #217 doc misstated the broker gate; attestation-level gap only (no plaintext moves; broker-presence still enforced by `NoCredentialBroker`). | **Fixed**: `docs/contracts/context-credential-resolution.md` now names both bases and records the caveat — an elevated run's attestation must not be keyed on the contract claim alone; the broker's provisioning check is the enforcing gate. Compile-time cannot know future elevations, so contract-level attestation for elevated runs is deferred to the materialization slice (where the contract shape must change anyway). |
| F-2 | P2 | `docs/contracts/storage.md` still said schema v10 after the v11 bump (missed lockstep artifact). | **Fixed**: bumped to v11. |
| F-3 | P3 | Client-supplied `decided_at` allows future-dated "pre-arming". | **Accepted with disposition**: owner-approved, journaled, sticky-revocable, and coverage can never exceed the bounded window; UI-time clamping noted as follow-up polish. |
| F-4 | P3 | Elevations for non-`UseCredential` permissions have no enforcement consumer — journaled approvals that license nothing. | **Fixed**: `decide_elevation` now refuses any permission except `UseCredential` until its enforcement point lands; the lifecycle test covers the refusal. |
| F-5 | P3 | The audit's decision arm re-derives attribution but not the ceiling/not-already-granted policy invariants. | **Accepted as optional hardening** — the journal is written only by the trusted host process and is append-only (update/delete triggers); noted for the audit-hardening pass. |
| F-6 | P3 | Test gaps: protocol-level op tests, decide replay-conflict, revoke-retry through the daemon, non-Running decide. | **Accepted as follow-ups** before the materialization slice; the lifecycle test covers every claim the PR makes. |

PR #513, opened 2026-09-10
Desktop shell slice 1: Tauri 2 application over a headless-proven controller (#54)

RepairYourTech:
First GUI shell slice for #54: the application surface exists as a **Tauri 2 desktop app** whose entire behavior lives in a headless, unit-tested controller — the architecture proof the workbench's Tauri acceptance builds on. Based on refreshed main 905c448e.

## What this does

**`crates/symbiote-desktop`** — a Tauri 2 workspace member with two layers:

- **`controller` (headless core, no Tauri types):** owns one daemon generation at a time plus the connected driver. It prepares the private environment (0700 state dir, 0600 operator config naming the sandbox launcher), spawns `symbioted`, sequences the first-release flow (open repository → describe task → start dispatch → run → completion evidence → worktree diff), persists the driver's journal positions to disk, and resumes a fresh process from exactly those positions — after a desktop restart OR a daemon crash (its crash path SIGKILLs exactly like a dead process; graceful `stop` uses the shutdown operation).
- **Tauri commands (thin wrappers):** `begin_session`, `start_demo`, `read_journal`, `finish_demo`, `journal_position`, `stop_session`. The minimal UI page calls them via the global Tauri IPC with a strict CSP (`default-src 'self'`), and states the fixture-transport boundary in the product: the demo's model turn is the operator's explicitly labeled fixture transport; a live one requires explicit authorization for credentials and billing — nothing is spent.

**The architecture proof is a checked-in test:** the same controller the commands call drives a REAL daemon through the whole first-release flow across a SIGKILL — asserting the restored journal position, `completion_requested` state (evidence, never verified completion), the worker's report, and the produced file's content on disk. Headless, so it runs in CI.

## Deliberately not here

- Bundled release packaging (sidecars, icons in bundles), real Wayland/X11 focus/composition/accessibility evidence, Preview isolation, and the strict TypeScript/React workbench — the UI is a minimal plain-HTML page; the next slice migrates it onto the workbench toolchain on top of the proven controller.
- The demo environment/paths are operator-supplied inputs; a polished onboarding flow is UI work on the same controller.

## Verification

- New test `desktop_controller_drives_the_first_release_flow_across_a_crash` passes (headless, real daemon, real sandboxed tool execution).
- Full local gauntlet green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked`.
- CI: the contracts workflow now installs the WebKitGTK stack (libwebkit2gtk-4.1-dev et al.) so both toolchain jobs compile the Tauri crate — the 1.85.0 job remains the MSRV proof.
- Supporting: `DemoOutcome`/`WorktreeEvidence` are serde-serializable for the IPC boundary; `DemoWorkflow` gains a graceful shutdown; the workflow crate gains serde.

## Standing gate (unchanged)


RepairYourTech:
## Independent review record — round 1 (head 7962f0f6, fixes in 0412fb8a)

Reviewer: independent local reviewer agent. Verdict **APPROVE — no P0, no P1**; the reviewer confirmed the diff matches the claims exactly, the checked-in test genuinely proves restart/resume across a real SIGKILL'd daemon, and reproduced the MSRV proof locally (`cargo check --workspace` under **1.85.0** compiles the full Tauri tree; no locked crate declares a higher rust-version). Full gauntlet green at head; both workflow e2e tests still pass.

### Findings and dispositions

| F-1 | P2 | `begin_generation` guarded on the child handle, not the connected driver — a daemon that died on its own reported success with a dead socket, and a connect failure left an unowned child. | **Fixed in 0412fb8a** — guard is now the connected workflow; the controller connects before owning the child, and a connect failure kills+waits the just-spawned daemon. |
| F-2 | P2 | `stop()` had an unbounded `wait()` (a stuck daemon would hang the UI's Mutex) and the `Drop` kill fallback was dead code. | **Fixed in 0412fb8a** — `stop()` bounds its wait (5s) and falls back to the crash path; `Drop` now genuinely owns the guarantee. |
| F-3 | P3 | Daemon exit status discarded on `DaemonExited`. | **Follow-up** — carry the status in the error. |
| F-4 | P3 | Daemon stdout/stderr nulled (no desktop log file yet). | **Follow-up** — log-file slice. |
| F-5 | P3 | create-then-chmod TOCTOU on the state dir/config. | **Accepted** — matches the e2e pattern; the only "secret" is the labeled fixture string; socket auth is peer-credential-based; harden with mode-at-creation later. |
| F-6 | P3 | Corrupt positions file silently bootstraps at 0. | **Accepted** — safe failure (idempotent journal re-read, command-id receipts replay); surface loudly later. |
| F-7 | P3 | `.lock().expect(...)` could poison permanently if a command ever panics. | **Follow-up** — `unwrap_or_else(PoisonError::into_inner)`. |
| F-8 | P3 | CSP `style-src 'unsafe-inline'` unnecessary (no inline styles). | **Fixed in 0412fb8a** — dropped; `default-src 'self'` posture otherwise confirmed correct. |
| F-9 | P3 | Fixed `/tmp` UI defaults (local-DoS/stale-socket polish). | **Follow-up** — per-user defaults. |
| F-10 | P3 | Test scratch cleanup only on success. | **Follow-up** — Drop guard for symmetry. |

RepairYourTech:
## #513 — review follow-ups closed (PR #520, merged `020c0cdd`)

The four follow-ups from this slice's review round are closed on main. Reviewed head: `f99baabe`. Merge commit: `020c0cdd0b060439cd04540f69eadaa807d98cc8`. PR: https://github.com/RepairYourTech/SymbioteIDE/pull/520 (independent review record: https://github.com/RepairYourTech/SymbioteIDE/pull/520#issuecomment-5626813293)

### What landed

1. **Daemon status in `DaemonExited`.** `begin_generation` reports the child's exit status (exit code or signal) in the typed error — an operator sees the daemon FAILED, not just that it "exited". Test: an unparseable operator config kills the daemon pre-socket; the surfaced status is non-empty.
2. **Corrupt positions file surfaced, never silently ignored.** A persisted `desktop-positions.json` that exists but does not parse is a typed `CorruptPositions` error — silently resetting it would discard the resume point the whole restart/resume story depends on. The failed generation REAPS its just-spawned daemon: the wait loop now reaps on every observation path (review P2 fixed — the `try_wait` error propagation had been the one early return that didn't), so no early return between spawn and ownership transfer can leak a child.
3. **Poison-proof session lock.** All six Tauri commands take the session through one `lock_session` helper that recovers from mutex poisoning (`into_inner`); zero `.expect("session lock")` remain. Soundness: the guarded value is the controller itself, whose `Drop` still bounds the daemon and whose `begin_generation` is idempotent across the states this crate's commands can leave (the doc states precisely that it cannot detect an externally dead daemon — pre-existing, unchanged).
4. **Drop guard test.** `a_controller_dropped_without_stop_leaves_no_daemon_behind`: a controller dropped without `stop()` — a panicking or closed UI — leaves no daemon accepting on the socket (the connection is the truth; the socket file may linger after an unclean exit).

The first-release flow test was refactored onto a per-test tagged environment (the old scratch path was pid-only — a parallel-collision risk); the `TestEnv` Drop removes the scratch on success and panic paths.

### Verification

- Reviewer (rebuilt daemon binaries, then targeted suites): `symbiote-desktop --test controller` 4 passed (three new + the refactored flow test), clippy + fmt clean; the corrupt-positions serde analysis (anything parsing is a genuine positions list; `[]` is legitimately restorable) and the lock-soundness audit came back clean.
- Local gauntlet on `f99baabe`: fmt clean, `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0, `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures). CI green on `f99baabe`.

### Honest scope notes

- The poison-recovery soundness argument assumes the workflow crate's methods are panic-free mid-operation — verified structurally for this crate's commands, not proven for the workflow crate (outside the diff).
- Two-project `positions()` restart remains unexercised (driver-restart proof covers one project); noted as headroom in the #519 record too.
- Live model-turn verification for BOTH runtimes still requires explicit user authorization for credentials and billing — nothing was spent; all execution here is fixture/offline.

PR #516, opened 2026-09-10
Both worker loops file elevation asks without granting (#269)

RepairYourTech:
Next #269 remainder slice from refreshed main `7174a9b2`: both worker loops now *emit* an elevation ask through the Host. The ask is journaled evidence. It never licenses anything. Only a later owner `decide_elevation` can grant a lease.

## What this does

**Native.** `WorkerTransports::resolve_native_credential` wraps broker resolve. A `use_credential_not_granted` refusal journals `ElevationRequested` (`UseCredential`) via `file_elevation_ask`, then returns the same typed refusal so the run cannot proceed. The ask never satisfies `active_elevation`.

**External.** The Codex driver still refuses every harness escalation. `ExternalSession::refused_approvals()` now exposes those refusals; `run_external_boxed` maps command/file/network-shaped refusals onto `ExecuteProcess` / `MutateStream` / `Network` and files one ask per distinct permission. Filing happens even if the turn later fails. The harness is still denied — the driver never grants.

**Host is the only journal writer.** Loops do not call `decide_elevation`. Command ids are `worker-elevation-{dispatch}-{permission}-{at}` so a retry of the same observed ask replays honestly.

## Honest non-claims

- Decide remains owner-only; this slice does not auto-grant.
- Still UseCredential-only on the decide path (no enforcement consumer for other permissions). Filing a `Network`/`ExecuteProcess`/`MutateStream` *ask* from the harness does not license those permissions.
- Skill/MCP/secret trust metadata, mid-run policy change semantics, two-harness demo, mixed staffing, Project isolation remain later.
- Live model-turn verification for BOTH runtimes still requires explicit user authorization for credentials and billing. Not attempted. Nothing spent.
- #269 stays open.

## Verification

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked` including `native_credential_refusal_files_an_ask_and_never_licenses` and `external_harness_refusal_files_an_ask_and_never_licenses`

No new spending, credential changes, or paid fallback. Do not silently use native API billing for external execution or Codex credentials for native.

### DIFF
--- a/crates/symbiote-external-agent/src/lib.rs
+++ b/crates/symbiote-external-agent/src/lib.rs
@@ -264,6 +264,7 @@ pub struct ExternalSession {
     harness_thread_id: Option<String>,
     turns_observed: u32,
     approvals_refused: u32,
+    refused: Vec<ApprovalRefusal>,
     usage_turns: u32,
     unreported_usage_turns: u32,
     completion_report: Option<String>,
@@ -298,6 +299,7 @@ impl ExternalSession {
             harness_thread_id: None,
             turns_observed: 0,
             approvals_refused: 0,
+            refused: Vec::new(),
             usage_turns: 0,
             unreported_usage_turns: 0,
             completion_report: None,
@@ -315,6 +317,13 @@ impl ExternalSession {
 
     pub fn harness_thread_id(&self) -> Option<&str> {
         self.harness_thread_id.as_deref()
+    }
+
+    /// The harness escalations this turn refused. Observation only: the
+    /// Host may journal an elevation *ask* from these; the driver never
+    /// grants.
+    pub fn refused_approvals(&self) -> &[ApprovalRefusal] {
+        &self.refused
     }
 
     pub fn run_summary(&self) -> ExternalRun {
@@ -567,6 +576,7 @@ impl ExternalSession {
             while let Some(request) = transport.recv_server_request()? {
                 let refusal = ApprovalRefusal::from_method(&request.method);
                 self.approvals_refused += 1;
+                self.refused.push(refusal);
                 let decision = if refusal.has_shaped_denial() {
                     ApprovalDecision::Denied
                 } else {
--- a/crates/symbiote-external-agent/src/lib.rs
+++ b/crates/symbiote-external-agent/src/lib.rs
@@ -319,11 +319,11 @@ impl ExternalSession {
         self.harness_thread_id.as_deref()
     }
 
-    /// The harness escalations this turn refused. Observation only: the
-    /// Host may journal an elevation *ask* from these; the driver never
-    /// grants.
-    pub fn refused_approvals(&self) -> &[ApprovalRefusal] {
-        &self.refused
+    /// The harness escalations this turn refused, drained so a later turn
+    /// cannot re-file earlier turns' evidence. Observation only: the Host
+    /// may journal an elevation *ask* from these; the driver never grants.
+    pub fn refused_approvals(&mut self) -> Vec<ApprovalRefusal> {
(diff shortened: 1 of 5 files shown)


RepairYourTech:
## Independent review record (local reviewer agent — posted by orchestrator)

An independent reviewer agent inspected `7174a9b2..2ce74aa7` hunk-by-hunk and runtime-verified the branch offline (no network, fixtures only).

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"The core invariant holds: no code path in this PR creates an `ElevationLease`, `use_credential_granted` is computed only from the binding grant or `active_elevation`, and the store's replay audit validates ask lineage (Running attribution, unique ids). Acceptance criteria met."*

### Findings and resolution

1. **P2 — derived ask id could overflow the 128-char `CommandId` bound** (`runner.rs`): a `DispatchId` may itself run to 128 chars, so `worker-elevation-{dispatch}-{permission}-{at}` could hard-fail the run exactly when it should file evidence of a refusal. **Fixed in 2d61e33a**: the id now carries a stable FNV-1a digest of the dispatch id (chosen over `DefaultHasher` so Host retries mint the same id across Rust versions and replay honestly instead of duplicating evidence), with permission and clock in the clear. Regression test `elevation_ask_id_stays_within_the_command_id_bound` pins the bound for a 128-char dispatch id at `Timestamp(u64::MAX)`.
2. **P3 — ask-filing failure could mask the turn error**: **fixed** — the ask is still attempted first (evidence preserved even on failed turns), but the turn error now outranks a filing failure so the operator sees why the run actually failed.
3. **P3 — `refused` accumulated across turns and was never drained**: **fixed** — `ExternalSession::refused_approvals()` now drains (`std::mem::take`) so a later turn cannot re-file earlier turns' evidence as duplicates.
4. **P3 — refusal matched by string literal** (`CredentialRefused("use_credential_not_granted")`): **deferred to the #269 remainder** — a typed variant would ripple into the service error mapping (`service.rs` maps `CredentialRefused`), which is wider than the reviewed diff; the name mapping is a single `broker_error_name` function.
5. **P3 (nit)** — runner tests assert the journal event and `active_elevation == false` but not a direct `elevation_requests` row count; the store's own `an_elevation_request_is_journaled_evidence_and_never_licenses` covers table readback, so no change.

Reviewer also verified as correct (no action): id-collision behavior (same ask twice at the same clock → identical journal bytes → honest replay; same id with different bytes → loud `IdempotencyConflict`), `MIGRATION_V12` creating only `elevation_requests` with ordered drop lists, and version lockstep (all non-16 minor fixtures are intentional rejected-version lists).

### Verification

- Reviewer, targeted suites on `2ce74aa7`: `symbiote-host --lib` 27 passed (incl. both `*_refusal_files_an_ask_and_never_licenses`), `symbiote-host --test daemon` 19 passed, `symbiote-store elevation` 2 passed, `symbiote-protocol --test contracts` 19 passed, `symbiote-external-agent` 29 passed, clippy on the five touched crates clean.
- Follow-up gauntlet on fixed head `2d61e33a`: `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures; all three elevation runner tests pass).

Reviewer-noted limits, unchanged by this PR: principal construction at the daemon transport boundary is outside the diff, and live external-harness (real Codex approval) semantics remain unverified — live model turns for **both** runtimes still require explicit user authorization for credentials and billing. All execution here is fixture/offline.

PR #518, opened 2026-09-10
Two-harness demo: native and external workers on one project without leakage (#269)

RepairYourTech:
Closes one slice of #269 (does not close the issue).

## What this does

The remaining #269 demo scope: **mixed staffing on ONE project** — a native worker and an external-harness worker, each with its own role, binding, objective, task, stream, and worktree, driven end to end through the real daemon over the wire, with the isolation between the lanes made observable and asserted.

**Operator-config fixture harness (product code).** `OperatorConfig` grows `external_fixture` (`thread_id`, `agent_message`): an explicitly labeled **scripted** Codex App-Server conversation served by the new `ScriptedCodexTransport`/`FixtureExternalFactory`, symmetric with the existing `native_fixture`. No pinned binary, no model turn, nothing spent. The harness conversation is scripted; everything the daemon does around it is real (provisioning, runtime-kind strictness, escalation refusal — none are scripted, completion-evidence filing). A live Codex turn remains gated on explicit user authorization for credentials and billing, and this fixture must never be presented as a model or as Codex.

**The Host advertises what it can actually execute.** The daemon's Host record now lists `ExternalHarness` under `supported_runtimes` ONLY when the operator provisioned the external execution path (`WorkerTransports::has_external_transport()`), mirroring the existing `Credentials`-claim rule. Without operator provisioning, a binding whose profile is `EXTERNAL_HARNESS` refuses at start — never a silent fallback to the native runtime, and never native billing for external execution.

**The demo driver gains the external lane.** `DemoWorkflow::start_demo_external` composes the second staffing lane on the same project (third role + Team member, binding with profile runtime `EXTERNAL_HARNESS` and its installation pinned, own objective/task/stream with seed-derived worktree identity). `finish_demo_external` runs and reads it. Both finish methods now (a) pin the runtime from the dispatch contract — a crossed lane returns an error, the caller cannot choose the runtime — and (b) filter completion reports per task, since the journal now carries two lanes.

**The e2e proof** (`two_harness_demo_..._without_leakage`): both lanes start on one project, the daemon is SIGKILLed exactly like a crashed desktop process, and BOTH dispatches resume and run on the restarted daemon. Assertions pin what "without leakage" means observably:
- Work isolation: each lane's completion report is its own worker's text (the journal filter is per task); the native worktree contains `produced.txt` (real sandboxed tool), the external worktree is a different directory with no native output in it.
- Configuration/credential isolation, read back over the wire via `get_binding`: the native binding grants `UseCredential` and references the broker-registered fixture credential; the external binding grants NO `UseCredential` and its profile references the harness's own credential — the operator's broker has no registration for it, so even the native arm's gate would refuse. Cross-project credential denial remains pinned by the existing broker tests.
- Runtime strictness: `WorkerRun.runtime` is `native_symbiote` for the native dispatch and `external_harness` for the external one — from each dispatch's contract.

## Scope bounds

- No protocol bump, no new wire operations (`get_binding` already existed).
- Worker completion remains evidence: both lanes end `completion_requested`; Host verification and independent review stay the gates.
- The demo proves both harnesses against the SAME control plane with fixture transports. Live model-turn verification for BOTH runtimes (native OpenAI Responses API, external Codex App Server) still requires explicit user authorization for credentials and billing — nothing was spent, no live turn attempted.
- Still open on #269: runtime-aware decide refusal, standalone MutateStream/Network consumers, skill/MCP/secret trust metadata, mid-run policy changes.

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures), including the new e2e plus the pre-existing first-release demo, driver-restart, and desktop-controller proofs with the extended Team.


RepairYourTech:
## Independent review record (local reviewer agent — posted by orchestrator)

An independent reviewer agent inspected `7322f435..3262b460` and adversarially verified the branch offline (no network, fixtures only; it rebuilt the daemon binary so the e2e suite exercised the PR's `symbioted`, not a stale one).

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"Acceptance criteria are met."* The reviewer verified adversarially that the no-fallback guarantee holds structurally (`Dispatch::compile` refuses `IneligibleHost` when the profile runtime is not in the live host record's `supported_runtimes`, and `start_prepared_task` recompiles with that record inside the transaction — no re-routing path exists); team validation passes with the third member; report filtering is safe across journal page boundaries (events accumulate across ALL pages before the per-task filter); the external arm never calls `resolve_native_credential` and even an elevation could only reach the unregistered `external-harness-ref` (`UnknownReference`), with profile-ref bounding preventing reach into `native-vault-ref`; the fixture choreography exactly matches the driver's handshake/correlation contract with `approvals_refused` staying 0; restart works for the second lane (dispatch ids deterministic per task, all state journaled); and the fixture is explicitly labeled everywhere with no live external factory in the tree.

### Findings and resolution

1. **P2 — the no-fallback claim had no pinning test**: **fixed in `bdbd8c25`** — new e2e `an_external_binding_refuses_to_start_without_operator_provisioned_harness_support`: with an operator config lacking `external_fixture`, the native lane still works and `start_demo_external` refuses AT THE DAEMON (never rerouted to native). Honest scope note surfaced while writing it: preparation is host-agnostic by design (it records `Ready`), so the refusal surfaces at start; the domain's `IneligibleHost` currently maps to the coarse `invalid_request` wire code, so the test pins the refusal (not a success) without enshrining the coarse mapping.
2. **P3 — duplicate predicate**: the dead `native_configured`/`external_configured`/`shell_configured` trio deleted; the operator-provisioning predicates are `has_credential_broker`/`has_external_transport`.
3. **P3 — leftover debug output + non-exhaustive body check in `start_demo_external`**: the stray `eprintln!` removed and the preparation body match made exhaustive (`_ => UnexpectedBody`).
4. **P3 — worktree-derivation logic duplicated a third time**: extracted `derive_stream_worktree`, shared by both lanes and kept in one place so the derivation cannot diverge from the evidence reader.
5. **P3 — lane coupling (external role registered inside the native half)**: accepted for the demo sequencer; the doc discloses the ordering (`start_demo` then `start_demo_external`).

Reviewer-noted limits (not verified from code alone): concurrent (simultaneous) execution of both lanes is not exercised — the demo runs them sequentially; lease-expiry timing across a slow restart is untested; "gated on explicit user authorization" for the live external path is currently vacuously true because no live external factory exists in the tree (the claim becomes auditable when one is introduced).

### Verification

- Reviewer, targeted suites on `3262b460` after rebuilding the daemon: `symbiote-workflow --test end_to_end` 3 passed, `symbiote-host --lib` 32 passed, `symbiote-external-agent --lib` 29 passed, `symbiote-domain --test contracts` 20 passed, `symbiote-store --lib` 55 passed, fmt + clippy clean.
- Follow-up gauntlet on `bdbd8c25`: `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures). One pre-existing load flake was observed and ruled out: `symbiote-sandbox process::setup_failure_stderr_is_bounded_explicit_and_debug_redacted` failed once in a parallel workspace run, passed in isolation, and passed in the full rerun — it exercises sandbox stderr bounds and touches none of this PR's code.

PR #520, opened 2026-09-10
Close the #513 review follow-ups: daemon status, corrupt positions, poison-proof lock, Drop guard test

RepairYourTech:
Closes the four review follow-ups left open by PR #513 (desktop shell slice).

## What this does

1. **Daemon status surfaced in `DaemonExited`.** `begin_generation` now reports the child's exit status (exit code or signal) in the error instead of discarding it — an operator sees the daemon FAILED and why-ish (e.g. `exit status: 1`), not just that it "exited".
2. **A corrupt positions file is surfaced, never silently ignored.** A persisted `desktop-positions.json` that exists but does not parse is a typed `CorruptPositions` error. Silently treating it as absent would reset the resume point the whole restart/resume story depends on — the operator decides (delete or restore), never the code behind their back. The failed generation also REAPS its just-spawned daemon: the previous error path would have leaked the child (it returned before the controller took ownership).
3. **Poison-proof session lock.** A panicking Tauri command poisoned the `Mutex<Option<DesktopController>>` and every later `.expect("session lock")` would brick the whole session. All six commands now take the session through one `lock_session` helper that recovers from poisoning — sound here because the guarded value is the controller itself, whose `Drop` still bounds the daemon and whose `begin_generation` is idempotent across whatever mid-operation state the unwinding command left.
4. **Drop guard test.** A controller dropped WITHOUT `stop()` — a panicking or closed UI, not a graceful shutdown — is proven to leave no daemon behind (connection to the socket fails after the drop; the socket file may linger, the connection is the truth).

## Tests

- `a_controller_dropped_without_stop_leaves_no_daemon_behind` — the Drop guard proof.
- `a_corrupt_positions_file_is_surfaced_not_silently_ignored` — typed error + reaped daemon.
- `begin_generation_reports_the_daemon_status_when_it_exits` — non-empty status surfaced (an unparseable operator config makes the daemon exit at startup).
- The pre-existing first-release flow test is refactored onto a per-test tagged environment so parallel runs never collide (the old scratch path was keyed by pid only).

## Scope bounds

- No protocol change, no daemon change — desktop crate only.
- The logged-out/default-lane behavior of the demo flow is unchanged; the workbench typecheck and CI desktop job cover the UI side.

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures), including the three new desktop tests plus the refactored first-release flow test.

### DIFF
--- a/crates/symbiote-desktop/src/controller.rs
+++ b/crates/symbiote-desktop/src/controller.rs
@@ -56,10 +56,15 @@ pub enum DesktopError {
     Setup(String),
     /// The workflow step failed; the typed error is preserved.
     Workflow(WorkflowError),
-    /// The daemon exited before its socket was ready.
-    DaemonExited,
+    /// The daemon exited before its socket was ready; the child's exit
+    /// status is surfaced (exit code or signal) instead of discarded.
+    DaemonExited(String),
     /// The daemon's socket never became ready.
     DaemonUnready,
+    /// The persisted journal positions file exists but does not parse.
+    /// Surfaced — never silently ignored: silently resetting it would
+    /// discard the resume point the whole restart story depends on.
+    CorruptPositions(String),
 }
 
 impl std::fmt::Display for DesktopError {
@@ -69,8 +74,16 @@ impl std::fmt::Display for DesktopError {
             Self::NotARepository => write!(f, "repository is not a git work tree"),
             Self::Setup(message) => write!(f, "desktop setup failed: {message}"),
             Self::Workflow(error) => write!(f, "{error}"),
-            Self::DaemonExited => write!(f, "daemon exited before becoming ready"),
+            Self::DaemonExited(status) => {
+                write!(f, "daemon exited before becoming ready: {status}")
+            }
             Self::DaemonUnready => write!(f, "daemon socket never became ready"),
+            Self::CorruptPositions(error) => {
+                write!(
+                    f,
+                    "persisted journal positions are corrupt (delete or restore the file, never silently reset): {error}"
+                )
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


RepairYourTech:

An independent reviewer agent inspected `d2a908a5..b818349e` and audited all four follow-ups offline.

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"This PR closes all four #513 review follow-ups."* The reviewer verified: the DaemonExited status premise is real (config load happens before serving, so the test's premise is genuine); the corrupt-positions surfacing is complete — `JournalPosition`'s serde is strict enough that anything parsing is a genuine positions list, and an empty list is legitimately restorable (persist writes exactly `[]` for a fresh session, semantically identical to a fresh boot); all six commands go through `lock_session` with zero `.expect("session lock")` remaining and exactly one `.lock()` site in the crate; the recovery is sound (commands don't mutate the controller before their fallible calls, `begin_session` doesn't hold the lock during `begin_generation`, and a mid-`stop_session` panic drops the controller exactly as before); the four tests ran concurrently, exercising the tag isolation.

1. **P2 — the `try_wait()` error path leaked the child**, the one early return between spawn and ownership transfer that didn't reap, contradicting the new reaping comment: **fixed** — the wait loop now reaps on every observation path (exited, wait-error), restructured so the audit invariant is structural, not incidental.
2. **P3 — `lock_session` doc overstated idempotency**: now states precisely that `begin_generation` is idempotent across the states this crate's commands can leave and cannot detect an externally dead daemon (pre-existing behavior, unchanged).
3. **P3 — `daemon_is_gone` semantics** (closed-listener vs. dead process): no change needed — in both users the child is provably reaped before the check, so a reaped process can never accept again; noted for future reuse of the helper.

Reviewer-noted limits: the poison-recovery soundness argument assumes the workflow crate's methods are panic-free mid-operation (verified structurally for this crate's commands, not proven for the workflow crate — outside this diff); the practical unreachability of the `try_wait` Err branch is an argument from std's internals, and the path now reaps regardless.

### Verification

- Reviewer: `symbiote-desktop --test controller` 4 passed (all three new tests + the refactored first-release flow test), clippy + fmt clean, after rebuilding the daemon binaries.

PR #517, opened 2026-09-10
A decided ExecuteProcess lease licenses the native shell path (#269)

RepairYourTech:
Closes one slice of #269 (does not close the issue).

## What this does

PR #516 wired both worker loops to *ask* for permissions they lack; `decide_elevation` still refused every non-`UseCredential` decision because nothing consumed the resulting lease. This slice lands the first non-credential enforcement consumer: the native shell-tool path.

**The consumer.** The sandbox refuses to launch unless the consent snapshot's access carries `ExecuteProcess` (`symbiote-sandbox` grant check), and that access is the Host-resolved current policy. New `runner::resolve_shell_access` resolves that policy: the dispatch binding's access, widened with `ExecuteProcess` ONLY while an owner-decided elevation lease for THIS dispatch is active at the run's clock — the same per-run basis as the broker's `UseCredential` gate. `service.rs` consumes it when building `ShellExecutorInputs`. The sandbox and trust-layer checks are unchanged: the consent's recorded access must still be a subset of the resolved policy, and the sandbox still re-checks every grant it requires (`ReadRoot`, `ExecuteProcess`, `MutateStream` for the write profile).

**The decide scope.** The store's consumer allowlist widens to `UseCredential` + `ExecuteProcess`. `MutateStream` and `Network` still refuse `InvalidElevation` (no consumer). Honest note: a Running dispatch's binding always carries `MutateStream` (`Dispatch::compile` requires it), so the never-re-grant check shadows the allowlist for `MutateStream` — pinned anyway in tests with that caveat stated.

**Never-license invariants pinned by tests.**
- `a_decided_execute_process_lease_licenses_the_shell_path_alone`: base policy refuses tools; a filed ASK does not widen; a decided lease widens EXACTLY `ExecuteProcess` (same roots, same revision, +1 grant); expiry is auto-revocation at the use site; sticky revocation kills it inside the window.
- `the_store_decides_only_permissions_with_enforcement_consumers`: `Network`/`MutateStream` decisions refuse; `ExecuteProcess` decides and `active_elevation` reports.
- `a_binding_execute_process_grant_needs_no_lease`: a binding that already grants it needs no elevation (and decides would refuse as re-grant).
[...]

PR #521, opened 2026-09-10
decide_elevation is runtime-aware: external dispatches refuse their leases (#269)

RepairYourTech:
Closes the deferred #517/#518 review item on #269 (does not close the issue).

## What this does

A decided lease licenses only Host-enforced gates, and **every such gate lives on the native loop**: the broker's `UseCredential` lease gate (#217) and the shell path's `ExecuteProcess` policy (#517). The external harness never consults `active_elevation` — its escalations are refused by the driver regardless of any lease (#516/#518). So an owner decision on an `EXTERNAL_HARNESS` dispatch — whichever permission it names — would journal an approval that licenses nothing and could mislead an auditor.

`decide_elevation` now refuses any decision whose Running dispatch's contract runtime is `EXTERNAL_HARNESS` as `InvalidElevation`, placed after the permission-consumer allowlist and before the re-grant/ceiling checks. Docs updated in both places that describe the decide scope (dispatch-preparation.md, protocol.md).

The rule is a generalization of the deferred review finding (ExecuteProcess on an external dispatch) to the honest full form: **UseCredential on an external dispatch licenses nothing too** — the external arm never resolves a native credential lease, so the broker's cross-project and gate protections could never be reached by such a lease anyway.

## The test

`a_decision_on_an_external_dispatch_refuses_its_lease_entirely`: an external dispatch whose binding LACKS both `ExecuteProcess` and `UseCredential` (so the refusals are unambiguously the runtime rule, not a re-grant accident), with the Team ceiling carrying both: decisions naming either permission refuse `InvalidElevation`, and `active_elevation` reports nothing afterwards. The existing native positives from #517 (decide → lease → shell policy widened; ask never widens) are unchanged and still pass.

## Scope bounds

- No protocol bump: `decide_elevation` (v1.15) wire shape unchanged; the refusal identity (`InvalidElevation` → `invalid_request`) is unchanged.
- Replay/audit untouched: the runtime rule is a write-path check, consistent with the permission allowlist (no journal exists that violates it — all prior decisions were on native dispatches).

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures).


RepairYourTech:

An independent reviewer agent inspected `020c0cdd..e072069b` and verified the rule offline.

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"The acceptance criteria are satisfied."* The reviewer verified: the test is genuinely unambiguous (binding grants = {ReadRoot, MutateStream} so the re-grant check cannot be the refuser; the fixture ceiling carries both permissions so the ceiling cannot be the refuser; both permissions are on the allowlist); the runtime is read from the same field path the run path itself gates on (`dispatch.contract().profile().runtime`, identical to `check_runnable` and service.rs), and the `Dispatch` aggregate's contract is non-optional with a custom `Deserialize` requiring `profile` — a hypothetical historical gap would fail loudly at deserialization, not silently bypass the rule; the journal audit re-derives only attribution, so the write-path-only rule matches how the permission allowlist is already handled and no journal on main can newly fail replay; the docs match the code exactly and the wire is genuinely unchanged (a pure write-path policy change, so amending the v1.15 description without a version bump is correct).

1. **P2 — fail-open for a future third `RuntimeKind`**: the check was equality against `ExternalHarness`, but the invariant is an allowlist statement. **Fixed**: expressed as `!= NativeSymbiote` so a future third runtime kind fails closed, exactly like the permission allowlist above it.
2. **P3 — stale store-test comment** ("this slice's only consumer is the broker's UseCredential gate", pre-existing on base): refreshed to the current consumer list.
3. **P3 noted, no change needed**: the test's ceiling claim was verified (ceiling = base {ReadRoot, ExecuteProcess, UseCredential} + MutateStream carries both permissions).

Reviewer-noted limit: whether any production journal outside this repo contains a pre-change external-dispatch decision cannot be inspected from here; within the repo, every decided-lease fixture and test site is native-runtime, so nothing breaks replay.

### Verification

- Reviewer, on `e072069b`: `symbiote-store elevation` 2 passed, `symbiote-store --lib` 55 passed, `symbiote-host --lib` 33 passed (incl. the new test), clippy + fmt clean.

PR #519, opened 2026-09-10
Project-isolation demo: two projects on one daemon without leakage (#54)

RepairYourTech:
Part of the first-release definition of done tracked via #54 (Project isolation); #269's issue evidence already cross-references this demo capability.

## What this does

**"Switch between two Projects without leaking work, configuration or credentials"** — demonstrated end to end over the wire: TWO projects on ONE daemon, each with its own repository, Team, bindings, task, and stream.

**The demo driver is now parameterized by project lane.** `DemoLane` carries a project's full canonical identity (project, root, objective, task, stream, lead/executor role ids, binding id, provider/model ids, credential reference). `start_demo_lane`/`finish_demo_lane` drive any lane; the existing `start_demo`/`finish_demo`/`read_journal` keep their signatures (default lane), so the desktop controller is untouched. Two new flows: `start_followon_lane` (additional tasks in an existing project, binding optional) and `replace_lane_binding` (the operator's correction: read the current binding, replace it under an exact revision CAS with a corrected credential reference — a dispatch compiled before the replacement keeps its snapshot; nothing running is silently rewired).

**The isolation e2e** (`two_projects_on_one_daemon_without_work_configuration_or_credential_leakage`) pins, over the wire with the real daemon:
- **Credential isolation**: a leak-attempt lane whose binding references the OTHER project's registered credential passes the `UseCredential` gate and is REFUSED by the broker's owning-project check — `credential lease refused: cross_project_denied`. The operator's revision-CAS correction fixes the reference, and the corrected lane runs to `completion_requested` with its own credential.
- **Work isolation**: project B's derived worktree does not exist before its own run (project A's run never created anything under B's namespace); the two projects' worktrees are different directories, each holding only its own sandboxed output.
- **Read scoping**: a foreign task id is `PermissionDenied` (task ids share one global namespace — cross-project existence is never revealed); a foreign binding id is `NotFound` ((project, id)-scoped).
- **Journal scoping**: both projects' evidence trails are paged fully; each carries none of the other's identities.

**Scoping facts the demo surfaced, now documented** (dispatch-preparation.md): role ids and root ids are globally unique across the daemon (register refuses duplicates by id alone); provider connections and model descriptors are Host-global records, so each lane names its own; command ids are journaled globally and the driver namespaces them per lane.

## Scope bounds

- No protocol bump, no new wire operations; the desktop controller is unchanged (default-lane API intact).
- Worker completion remains evidence: both projects' lanes end `completion_requested`; Host verification and independent review stay the gates.
[...]

PR #522, opened 2026-09-11
Positions restart tracks two projects and both lanes across a crash (#54)

RepairYourTech:
Closes the positions-restart test-coverage headroom that the independent reviews of #518 and #519 both named.

## What this does

The first-release story says a desktop restart resumes without losing canonical state — and since #519 the desktop can track TWO projects. This PR proves the driver's per-project journal positions survive a daemon SIGKILL in that full form: a driver tracking **two projects** (the staffing project with its native lane, plus a second project with the broker-refused leak lane, the operator's revision-CAS binding correction, and the corrected lane) serializes its positions to disk exactly as a desktop shell would, the daemon is SIGKILLed, and a **fresh driver reconstructed from the serialized file** — sharing no memory with the dead one — resumes with BOTH projects' cursors restored exactly, then finishes both lanes on the restarted daemon.

Assertions: both projects' pre-kill cursors > 0; both restored cursors equal the pre-kill values exactly; both lanes end `completion_requested`; each lane's report is its own worker's text via the per-task journal filter; the projects' worktrees are distinct directories.

## Honest notes

- Both native lanes read the ONE global fixture factory, so the report *texts* are equal by fixture design — the test says so explicitly and pins isolation through the per-task journal filter and distinct worktrees instead (the Project-isolation test additionally proves the worktree/credential boundaries).
- The extracted lane helpers (`b_leak_lane`/`b_ok_lane`) deduplicate the Project-isolation test's definitions; no behavior change there.

## Scope bounds

- Test-only change: no protocol bump, no product code change, desktop controller untouched.

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures), including the new test and all pre-existing demos.

### DIFF
--- a/crates/symbiote-workflow/tests/end_to_end.rs
+++ b/crates/symbiote-workflow/tests/end_to_end.rs
@@ -813,7 +813,8 @@ fn positions_track_two_projects_and_both_lanes_across_a_daemon_crash() {
         .expect("lane B (ok) start");
     let a_cursor = workflow.read_journal().expect("journal A read");
     let b_cursor = workflow
-        .read_journal_lane(&b_leak_lane())
+        // Both lanes share project B; the lane choice only names the project.
+        .read_journal_lane(&b_ok_lane())
         .expect("journal B read");
     assert!(a_cursor > 0, "project A's evidence trail advanced");
     assert!(b_cursor > 0, "project B's evidence trail advanced");
@@ -830,6 +831,10 @@ fn positions_track_two_projects_and_both_lanes_across_a_daemon_crash() {
     let restored: Vec<symbiote_client_sdk::JournalPosition> =
         serde_json::from_slice(&std::fs::read(&positions_path).unwrap())
             .expect("restore serialized positions");
+    // Both projects are tracked: a restore that lost either key would
+    // default that project's cursor to 0 and fail the exact-equality
+    // asserts below — this pins the file's SHAPE too.
+    assert_eq!(restored.len(), 2, "one position per tracked project");
     let mut workflow =
         symbiote_workflow::DemoWorkflow::connect(state_dir, reservation_base, &env.repo)
             .expect("connect")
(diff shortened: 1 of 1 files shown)


RepairYourTech:

An independent reviewer agent inspected `09616d06..2a7b45b4` and verified the change offline (rebuilt the daemon binaries first).

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"It satisfies the acceptance criteria, and the cursor assertion is genuinely load-bearing"* — `journal_position` defaults to 0 for an untracked/mis-keyed project, so an empty or wrongly-keyed restore fails the exact-equality asserts; the two project keys cannot collide (`ProjectId` is a value-based newtype); and generation 2's driver is provably fresh (fresh `ClientSession`, positions through a serde/file round-trip, dead daemon a separate SIGKILLed process). The reviewer also verified the fixture-factory claim (one `echo_text` per operator config, cloned per run — equal report texts across native lanes are fixture design, not leakage), the per-task journal filter driving both finish calls, and that the diff is test-only.

1. **P3 — comment misdirection** (pointed at the Project-isolation test while the distinct-worktree assert sits right below): reworded to name both.
2. **P3 — one keying-regression shape could slip through if both cursors happened to be equal**: the serialized positions file's SHAPE is now pinned (`assert_eq!(restored.len(), 2)` — one entry per tracked project), so a restore that lost a key fails on shape before the cursor asserts.
3. **P3 — `read_journal_lane` used the leak lane solely for its project field**: now names the corrected lane with a comment (behaviorally identical — the call only reads `lane.project`).

Reviewer-noted limits: `a_cursor != b_cursor` in practice was not enumerated (the shape assert above makes the test robust regardless); SIGKILL-restart timing verified with two clean runs on this machine, not under heavy CI load.

### Verification

- Reviewer: `symbiote-workflow --test end_to_end` 6 passed, `symbiote-client-sdk` 8 passed (pins the restore/default/keying semantics the test relies on), clippy + fmt clean.

PR #524, opened 2026-09-11
Desktop Preview surface: untrusted worker output with no IPC (#54)

RepairYourTech:
Part of #54: before "secure Preview isolation" can be verified, the desktop needs a Preview surface at all — built with the isolation IN, not bolted on.

## What this does

**The Preview window is the untrusted surface.** After a run, the main window's new `Preview run output` button opens a separate Preview window rendering the run's output — the worker's completion report, the worktree path, and the produced/changed file names. The security properties are structural:

- **No Tauri IPC**: the window is labeled `preview` and named in NO capability file (`capabilities/default.json` lists only `windows: ["main"]`), so worker content shown there can never invoke the owner's commands or reach the local-owner policy.
- **Inert by construction**: the document is built by a pure Rust function (`preview_document`) that HTML-escapes every dynamic value, includes no script, and carries `Content-Security-Policy: default-src 'none'` both as a response header and a document meta. Worker content can only render as text, whatever it contains — a report of `<script>alert(1)</script>` renders as literal text.
- **Served through a dedicated custom protocol**: `preview://` serves exactly that managed document — no app assets, no filesystem paths, one CSP header. Re-opening refreshes in place via an owner-initiated reload navigation.
- The preview page itself states the isolation property in-product ("no application commands: worker content can never invoke the owner's operations").

## What this does NOT do (honest scope)

- No file *content* preview yet — the surface lists produced/changed file NAMES; bounded file-content and diff reading is the next increment.
- The live isolation verification (launch on the Wayland session, drive the run, open the Preview window, assert the window has no Tauri globals and a hostile report renders inert) is the NEXT slice — this slice's headless tests pin the escaping and document shape; per the working rules, the shell is not declared complete from headless tests alone.
- The live model-turn gate is unchanged: live native OpenAI Responses API turns and live Codex App Server turns require explicit user authorization for credentials and billing — nothing was spent; the surface renders the labeled fixture run's output.

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok), including two new headless tests (`worker_content_renders_as_inert_text`, `an_empty_outcome_still_carries_the_isolation_statement`).
- The committed `app/dist` is rebuilt (the CI desktop-workbench job rebuilds + typechecks + freshness-checks it).


RepairYourTech:

An independent reviewer agent reviewed the PR against the exact resolved Tauri crate sources (tauri-2.11.5, tauri-utils-2.9.3, tauri-build-2.6.3, wry-0.55.1) and the crate's freshly generated ACL artifacts.

**Verdict: REQUEST_CHANGES — all findings addressed in the follow-up commit.**

### The P1 and its fix

The reviewer's finding: **the headline "no Tauri IPC by construction" claim was false as configured.** Tauri's dispatch-level capability gate never applies to app commands unless the app defines an ACL manifest (`tauri_build::Attributes::app_manifest`); this crate defined none, so a window labeled `preview` — unlisted in capabilities — could still dispatch the owner's commands (`begin_session`, `finish_demo`, …). The reviewer verified the full chain in the vendored sources (the dispatch gate at `webview/mod.rs:1823` activates only with `has_app_acl_manifest`; the generated `acl-manifests.json` contained only `core*` keys; `preview://` is a local origin on Linux so the gate would have applied had it existed) and noted it was **not exploitable today** (worker content cannot execute: everything is HTML-escaped and the CSP denies script) — but the third advertised isolation layer was absent, and a future escaping/CSP regression would have silently exposed owner commands.

**Fix**: `build.rs` now defines the app ACL manifest for all seven commands, and `capabilities/default.json` grants the generated `allow-*` permissions to `main` only. The gate is now active, and the omission-based isolation is real by construction. The property is CI-pinned headlessly: build.rs exposes the generated `acl-manifests.json` and resolved `capabilities.json` via compile-time env, and a new test asserts the manifest exists (gate active), all seven generated `allow-*` permissions exist (identifier form: tauri-build hyphenates), and every resolved capability grants `main` and never names `preview`. The doc comments and the in-document statement now describe a property the code actually provides.

### P3s (all fixed)

- The protocol handler returns a 500 instead of `expect`-panicking across the GTK FFI boundary.
- A create-vs-create race falls back to the refresh path instead of failing the click.
- The document renders the dispatch id (a stale preview is never indistinguishable from the current run) and the UI clears the outcome on begin/stop.
- CSP hardened (`form-action 'none'; frame-ancestors 'none'; base-uri 'none'` + `X-Content-Type-Options: nosniff`, `Cache-Control: no-store`).
- A comment records that the `preview://` scheme is app-global (any window may navigate to it; the document is inert and only the Preview window points there).

### Verification

- Reviewer: `symbiote-desktop --lib` 2 passed (escaping tests), clippy + fmt clean, `app/dist` byte-identical after rebuild, strict TS + app tests pass.
- Follow-up gauntlet on the fixed head: fmt clean, `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0, `cargo test --workspace --locked` exit 0 (76 suite summaries ok) — including the new ACL pin test (now 3 lib tests).

Reviewer-noted limits: the live Wayland/WebKit behavior of the surface (header-CSP application to custom schemes, reload re-invoking the handler, a live dispatch from the preview window being rejected) is explicitly the NEXT slice's acceptance run on the Wayland session. The live model-turn gate is unchanged: **live native OpenAI Responses API turns and live Codex App Server turns require explicit user authorization for credentials and billing** — nothing was spent; the surface renders the labeled fixture run's output.

PR #523, opened 2026-09-11
Tauri desktop shell: real Wayland acceptance evidence (#54)

RepairYourTech:
Part of the first-release definition of done tracked via #54: **real Tauri acceptance on Wayland**.

## What this is

Acceptance evidence from the REAL binary (`target/debug/symbiote-desktop`, Tauri 2 + WebKitGTK) launched on the operator's live KDE/kwin **Wayland** session — not a headless test. The record: [docs/proofs/tauri-wayland-acceptance.md](docs/proofs/tauri-wayland-acceptance.md), with screenshots and the AT-SPI/KWin driver scripts under [`docs/proofs/evidence/tauri-wayland/`](docs/proofs/evidence/tauri-wayland/).

Captured (2026-09-10):
- **Launch**: the window maps on the kwin Wayland session; the GTK frame exposes itself as AT-SPI application `symbiote-desktop`.
- **Accessibility**: WebKitGTK exposes the whole React UI through AT-SPI — heading, the live-turn fixture notice, three labeled entries with values, every button (tree dumped with the committed `atspi_drive.py`).
- **Focus**: KWin scripting activates the window; the AT-SPI frame reports `STATE_ACTIVE`/`STATE_SHOWING`.
- **Composition**: `spectacle` captures the focused window fully rendered (not blank) — WebKitGTK composition works under Wayland.
- **Whole demo driven through the a11y tree**: `begin: started` (the app spawned the real `symbioted`) → `start-demo: dispatch disp_staffing-task` → `read-journal: 12` → `finish-demo: state=completion_requested`, and the run's sandboxed shell tool really wrote `produced.txt` inside the derived reserved worktree (filesystem-observed).
- **Graceful stop**: the UI `Stop` button — `stop: stopped`, daemon gone.

## Honest boundary

- One kwin/Wayland session on one machine; an XWayland/X11-run instance was not separately exercised; Windows/macOS remain explicitly pending.
- This UI has no Preview surface yet, so no Preview isolation evidence exists (separate definition-of-done item). No whole-process performance measurements taken.
- One observation: the app did not exit on `SIGTERM` (required SIGKILL after the session was already stopped); the graceful `Stop` path worked and left no daemon behind, and the #520 Drop guard bounds the non-graceful path. Noted as a follow-up, not claimed as a defect fix.
[...]

PR #525, opened 2026-09-11
Preview isolation verified live: hostile report renders inert, app ACL rejects preview invokes (#54)

RepairYourTech:
Closes the live-verification slice for the Preview surface built in #524 (the "secure Preview verification" definition-of-done item, Wayland part).

## What this does

1. **Hostile-by-default fixture report** (product change): the labeled fixture transport's report is now deliberately hostile markup — `<script>` appending an observable marker (`WORKER_SCRIPT_EXECUTED`), an `onerror` image, bold tags — defined once (`symbiote_workflow::demo::FIXTURE_REPORT`) and used by every fixture config and assertion. Every run of every demo now permanently proves that worker content renders as inert text on every surface.
2. **The Preview self-attests its isolation** (product change): after every page load, an owner-injected constant probe runs in the Preview window's own context, attempts the read-only owner command `journal_position` FROM THE PREVIEW, and reports the platform's verdict in the window title and the page body. Worker content never reaches the probe — it is a constant of the binary. The user sees the verdict every time they open the surface.
3. **Live verification** (docs/proofs/preview-isolation-acceptance.md + artifacts): the whole flow driven by AT-SPI on the real UI over the operator's kwin Wayland session; the Preview page shows the hostile markup as literal text with the execution marker appearing **exactly once** (inside the escaped literal — execution would append a second, standalone marker; the checker enforces this), and the probe's live verdict records the ACL rejection with the platform's own diagnostic: `journal_position not allowed on window "preview", webview "preview" … referenced by: capability: default, permission: allow-journal-position`.

## Honest scope

- No pixel capture of the Preview window: it sat behind the operator's active work and focus-stealing was avoided deliberately. The isolation proof does not need pixels — the AT-SPI capture reads the live DOM text, stronger than pixels for this claim. Rendering is evidenced by #523's committed screenshots.
- File-content/diff preview remains the next increment; skill/MCP/secret trust metadata and mid-run policy semantics remain on #269.
- **The live-turn gate is unchanged**: one live native OpenAI Responses API turn and one live Codex 0.118.0 App Server turn require explicit user authorization for credentials and billing — nothing was spent; all execution is fixture/offline.

## Verification

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures) — all workflow/desktop assertions updated to the hostile report constant.
- Live run: all six e2e-passing behaviors exercised through the real UI; evidence artifacts committed under docs/proofs/evidence/preview-isolation/.


RepairYourTech:

An independent reviewer agent reviewed `e545e51e..b5a32386` against the codebase and the committed live-run artifacts (rebuilding the spawned binaries first).

**Verdict: APPROVE_WITH_COMMENTS.** Summary from the review: *"It satisfies all three acceptance criteria."* The reviewer verified the hostile markup's blast radius is fully escaped on every surface it could find (React text nodes with `JSON.stringify` quoting; Rust `escape()` with ampersand-first ordering plus the deny-all CSP; serde-JSON journal paths — no `innerHTML`/`dangerouslySetInnerHTML`/`document.write` anywhere in the workspace, and the only `eval` call sites take compile-time constants), that no stale assertion on the old report text remains (the external lane's intentionally distinct fixture text is correct design — it preserves the two-harness report-isolation discrimination), and that the committed evidence matches the code and artifacts (`WORKER_SCRIPT_EXECUTED` exactly once in both AT-SPI captures, inside the escaped literal; the probe paragraph shows the platform's genuine ACL diagnostic naming `capability: default, permission: allow-journal-position`).

### Findings and resolution (fixed in `30b349a5` + `f19109a`-lineage commits on this branch)

1. **P3 — the probe labeled ANY invoke error as an ACL rejection**: the catch branch now distinguishes — only the ACL signature (`not allowed`) yields the OK verdict; anything else reports `UNVERIFIED: invoke rejected but not by the ACL`, so a future refactor cannot silently mask drift.
2. **P3 — the verdict paragraph was spoofable by report content**: the probe's verdict line is now bound to an **owner-generated nonce** (rendered in the document heading, unpredictable to worker content at report-generation time), so a report-embedded fake verdict line cannot impersonate the self-check. The document heading displays the nonce for matching.
3. **P3 — the window-title half of the verdict was not evidenced**: the record now states precisely that the verdict is evidenced in the page body (committed artifact) and that the title mutation is not captured.
4. **P3 — the FAILED/no-IPC probe branches are untested code**: noted honestly — the structural ACL property remains CI-pinned by `preview_ipc_tests`; a node-level probe-branch test is future polish.
5. **P3 — the record referenced a "checker" that was not committed**: the marker-count execution discriminator is now a committed `marker` mode of `atspi_drive.py` (the reproduce section references it), the dead `tree` stub is removed, and the unused action-description read is gone.

### Verification

- Reviewer: `symbiote-workflow --test end_to_end` 6 passed (against the hostile report), `symbiote-desktop --lib` 3 passed + `--test controller` 4 passed, clippy + fmt clean, committed `dist` byte-identical to a fresh rebuild, app tests 2/2.
- Follow-up gauntlet on the final head: `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` exit 0; `cargo test --workspace --locked` exit 0 (76 suite summaries ok, zero failures).
- The live artifacts were regenerated after the probe change: the committed capture carries the nonce-bound verdict `Isolation self-check [probe probe-3049777-403963934]: OK: app command rejected by the app ACL (journal_position not allowed on window "preview", … permission: allow-journal-position)`.

Standing gate, unchanged: **live model-turn verification for both runtimes (native OpenAI Responses API turn, live Codex 0.118.0 App Server turn) requires explicit user authorization for credentials and billing.** Nothing was spent; all execution here is fixture/offline.

PR #526, opened 2026-09-11
feat(preview): bounded run diff + untracked content heads on the Preview surface (#54)

RepairYourTech:
## What this changes

The Preview surface gains bounded diff evidence for a finished run, gathered owner-side (never from the front end).

**`symbiote-repo::observe_run_diff`** (new) — a `RunDiff` built from an already-observed `WorktreeStatus`. Collection is **total**: no tracked-side problem can cost the caller its untracked evidence, and every degradation is explicit.

- **tracked**: `git diff HEAD` (unified). Past the git output bound it degrades to `git diff HEAD --stat` with `tracked_truncated = true`; any other failure (a non-UTF-8 patch from a binary/latin-1 tracked file, a `--stat` summary that also exceeds the bound, a git refusal) becomes a bracketed placeholder plus a `tracked_note` naming the reason. An unborn HEAD is an empty tracked diff — not a refusal — and a HEAD that resolves to nothing while other refs exist is a broken HEAD and is named.
- **untracked**: per-path content heads, each at most 16 KiB, at most 32 paths, each carrying its own `truncated` flag; `untracked_omitted` counts the paths the cap left head-less.
- **never read**: the status path list is an advertisement, not authority. Non-relative paths (joining an absolute path *replaces* the base), `..` components and the C-quoted porcelain form are refused; the parent directory is canonicalized and must land **inside the worktree**, so an intermediate symlinked directory cannot walk the read out; and the final component is **opened** with `O_NOFOLLOW | O_NONBLOCK` and classified by the opened handle's own metadata — there is no check-then-open window, a symlinked final component is refused by the kernel, and a FIFO is neither waited on nor read.
- A byte bound cutting a multi-byte character keeps the complete character prefix instead of reporting a UTF-8 file as `<non-utf8 content>`.
- Tests drive real repositories: tracked + untracked, 16 KiB truncation, `--stat` degradation and its shape, symlink (final and parent), FIFO, absolute path, non-UTF-8 tracked patch, non-UTF-8 untracked content, broken HEAD, unborn HEAD, 2- and 3-byte boundary cuts, interior quotes, and the cap/omit count.

**Controller `run_diff(worktree)`** — canonicalizes both the requested path and the session's reservation base and refuses anything that does not resolve inside the base (component-wise, not string-prefix). This is structural: the front end's path string is never trusted. Tests cover a real repo inside the base, a repo outside it, a symlink inside pointing outside, and a name-prefix sibling (`worktreesEVIL`).

**`open_preview`** — gathers the diff through the controller and hands it to `preview_document`, which renders a *Tracked diff* section (with any degradation note) plus per-file untracked content heads through the same HTML escaper as every other dynamic value. A gathering refusal (outside the base, or no live session) degrades to an escaped owner note: the run's report still previews. Hostile content in a diff, a file path, or a file head renders inert — pinned by tests.

## Bounds, all explicit

| layer | bound | flag |
| --- | --- | --- |
| tracked diff output | git invocation output bound | `tracked_truncated` (the degradation is `--stat`); any other tracked failure → `tracked_note` |
| per-file untracked head | 16 KiB | `truncated` (per head) |
| untracked paths given heads | 32 | `untracked_omitted` |

## Security boundary

- The UI cannot widen the observation: the path must resolve inside the session's reservation base, and refusals are typed.
- Content is never read through a symlink — final or intermediate — and containment is checked on resolved paths.
- Everything rendered passes through `escape()`; the document keeps its `default-src 'none'` CSP and contains no script.

## Verification (head f09e347a, local, real exit codes)

```
cargo fmt --all --check                                          -> 0
cargo clippy --workspace --all-targets --locked -- -D warnings   -> 0
cargo test --workspace --locked                                  -> 0  (458 passed / 0 failed)
```

Plus the documented rebuild-first step (`cargo build -p symbiote-host --bins -p symbiote-sandbox --bin symbiote-sandbox-launch`) because e2e tests spawn the workspace `target/debug` daemon binary.

## Honest non-claims

- **No live GUI verification of this increment yet.** The new sections inherit the Preview surface's CSP and escaper, and the surface's no-IPC isolation was live-verified on Wayland in #525, but the diff rendering itself is pinned headlessly only — there is no screenshot or a11y capture of it. A live capture is a separate follow-up.
- The race-winning two-thread TOCTOU is not deterministically testable; the window is closed by construction (open with `O_NOFOLLOW`, classify the opened handle), and the deterministic escapes are pinned by tests.
- **Live model turns remain gated.** One live native OpenAI Responses API turn and one live Codex 0.118.0 App Server turn require explicit user authorization for credentials and billing. Nothing was spent; no live turn was attempted; the explicitly labeled fixture transports remain the only execution path.
- No new spending, no credential changes, and no paid fallback authorization is implied.

## Review status

Independent review (fresh local agent, headless, against `e72d674d`) returned **APPROVE_WITH_COMMENTS** with one P1 and three P2s; the full findings table, evidence, and disposition per finding is here: [#526 review record](https://github.com/RepairYourTech/SymbioteIDE/pull/526#issuecomment-5630762195). All P1/P2 findings are fixed in `f09e347a` with tests pinning each one; the single accepted P3 (a path exactly equal to the reservation base) is documented in that record. CodeRabbit reports "manual review required for this OSS repository" and Qodo is paused — both are stated as a coverage gap rather than counted as review.

### DIFF
--- a/crates/symbiote-desktop/src/controller.rs
+++ b/crates/symbiote-desktop/src/controller.rs
@@ -305,6 +305,27 @@ impl DesktopController {
         self.workflow = None;
     }
 
+    /// Bounded diff evidence for a finished run's worktree. The
+    /// requested path must resolve inside THIS controller's reservation
+    /// base — the UI is the owner surface, but the boundary stays
+    /// structural rather than trusting the front end's path strings.
+    pub fn run_diff(&self, worktree: &str) -> Result<symbiote_repo::RunDiff, DesktopError> {
+        let requested = std::fs::canonicalize(worktree)
+            .map_err(|error| DesktopError::Setup(format!("worktree resolve: {error}")))?;
+        let base = std::fs::canonicalize(&self.reservation_base)
+            .map_err(|error| DesktopError::Setup(format!("reservation base: {error}")))?;
+        if !requested.starts_with(&base) {
+            return Err(DesktopError::Setup(
+                "worktree outside this session's reservation base".into(),
+            ));
+        }
+        let mut git = symbiote_repo::SystemGit::new();
+        let status = symbiote_repo::observe_status(&mut git, &requested)
+            .map_err(|error| DesktopError::Setup(error.to_string()))?;
+        symbiote_repo::observe_run_diff(&mut git, &requested, &status)
+            .map_err(|error| DesktopError::Setup(error.to_string()))
+    }
+
     fn workflow(&mut self) -> Result<&mut DemoWorkflow, DesktopError> {
         self.workflow.as_mut().ok_or(DesktopError::Setup(
             "no connected generation; begin one".into(),
--- a/crates/symbiote-desktop/src/controller.rs
+++ b/crates/symbiote-desktop/src/controller.rs
@@ -322,8 +322,12 @@ impl DesktopController {
         let mut git = symbiote_repo::SystemGit::new();
         let status = symbiote_repo::observe_status(&mut git, &requested)
             .map_err(|error| DesktopError::Setup(error.to_string()))?;
-        symbiote_repo::observe_run_diff(&mut git, &requested, &status)
-            .map_err(|error| DesktopError::Setup(error.to_string()))
+        // Collecting the diff itself is total: every degradation is inside
+        // the returned RunDiff (flags and notes), so a partly unreadable
+        // worktree still previews what it can.
+        Ok(symbiote_repo::observe_run_diff(
+            &mut git, &requested, &status,
+        ))
     }
 
     fn workflow(&mut self) -> Result<&mut DemoWorkflow, DesktopError> {
--- a/crates/symbiote-desktop/src/lib.rs
+++ b/crates/symbiote-desktop/src/lib.rs
@@ -43,6 +43,7 @@ fn preview_document(
     worktree: &str,
     files: &[String],
     diff: &symbiote_repo::RunDiff,
+    diff_note: &str,
 ) -> String {
     fn escape(value: &str) -> String {
         value
@@ -79,8 +80,15 @@ fn preview_document(
     document.push_str(&escape(worktree));
     document.push_str("</pre>\n");
     // The run's diff evidence — bounded upstream (symbiote-repo), escaped
-    // here like every other dynamic value.
+    // here like every other dynamic value. A gathering refusal is a note
+    // (owner-built text, escaped like everything else), so the report
+    // above still previews.
     document.push_str("<h2>Tracked diff</h2>\n");
+    if !diff_note.is_empty() {
+        document.push_str("<p>");
+        document.push_str(&escape(diff_note));
+        document.push_str("</p>\n");
+    }
     if diff.tracked_truncated {
         document.push_str("<p>The full diff exceeded the bound; this is the --stat summary.</p>\n");
     }
@@ -97,6 +105,15 @@ fn preview_document(
         document.push_str("</p>\n<pre>");
         document.push_str(&escape(&head.content));
         document.push_str("</pre>\n");
+    }
+    // The content-head cap is stated, not silent: the count below is how
+    // many untracked paths the bound left head-less.
+    if diff.untracked_omitted > 0 {
+        document.push_str("<p>");
+        document.push_str(&diff.untracked_omitted.to_string());
+        document.push_str(" further untracked path(s): content heads omitted (bound ");
+        document.push_str(&symbiote_repo::MAX_UNTRACKED_FILES.to_string());
+        document.push_str(")</p>\n");
     }
     document.push_str("<h2>Files produced (");
     document.push_str(&files.len().to_string());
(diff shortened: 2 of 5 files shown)


RepairYourTech:
## Independent review record — round 1 (verdict: APPROVE_WITH_COMMENTS)

**Reviewer**: a fresh local agent with no prior context on this branch, run headless against the exact reviewed head `e72d674d` (worktree `issue-54-preview-diff`, diff `main...HEAD`). It never wrote to the repository (`git status --porcelain` empty at the end), built its own adversarial probe crate under `/tmp/issue-54-probe` that path-depends on `symbiote-repo`, and drove the real methods. Local agents leave no GitHub trace, so this comment is the review record.

**Coverage gap, stated not papered over**: CodeRabbit reports `Review skipped: manual review required for this OSS repository` on this PR, and Qodo is paused (produces no review). The agent review below is the sole substantive review.

### Findings and disposition

| # | Sev | Location | Finding | Reviewer's empirical evidence | Disposition |
| --- | --- | --- | --- | --- | --- |
| 1 | **P1** | `symbiote-repo/src/lib.rs` `untracked_content` | The path list was trusted: `worktree.join(path)` **replaces** the base for an absolute path, and an intermediate symlinked directory was followed by both the metadata check and the open. Only the final component was protected. | Crafted `WorktreeStatus{untracked:["/etc/hostname"]}` returned this host's hostname; `link/secret.txt` with `link` → outside dir returned `OUTSIDE-SECRET\n`. Confirmed `git status --porcelain -z` itself never emits such paths (default and `showUntrackedFiles=all`), so not exploitable through the controller today — but the function's documented invariant ("never read a symlink, never leave the worktree") was false at the API. | **FIXED** in `f09e347a`: parent canonicalized and required inside the worktree, non-relative / `..` / C-quoted paths refused, and the final component **opened** with `O_NOFOLLOW`\|`O_NONBLOCK` and classified by the opened handle's metadata. Tests: `an_absolute_untracked_path_is_never_read`, `a_symlinked_parent_directory_is_never_followed`, `symlinked_untracked_paths_are_never_read`. |
| 2 | P2 | `head_is_unborn` | Every `rev-parse` refusal was treated as an unborn branch, so a broken HEAD previewed as "nothing changed" instead of a named failure. | `not_a_repo_run_diff_empty_status` → `SWALLOWED_AS_EMPTY`; HEAD repointed at a nonexistent ref → `tracked_len=0`, no error. | **FIXED**: the rule now mirrors `observe_head` (symbolic-ref resolves, `rev-parse` refuses) **and** asks `show-ref --quiet` whether any ref exists — nothing reachable from HEAD while other refs exist is a refusal, and it is named. Test: `a_broken_head_is_named_rather_than_reported_as_empty`. |
| 3 | P2 | `observe_run_diff` | Any tracked-side failure returned before the untracked heads were collected: one bad byte in a tracked file, or a `--stat` summary past the bound, wiped the entire diff section. | Tracked file with `0xff` and no NUL → `ERR MalformedOutput`; 1271 long-named files → `--stat` `ERR OutputTooLarge`. | **FIXED**: collection is now **total** (`observe_run_diff` returns `RunDiff`, not `Result`) — untracked heads are collected first, and tracked degradations are recorded in a new `tracked_note`. Tests: `a_non_utf8_tracked_diff_is_named_and_untracked_heads_survive`, `a_tracked_degradation_note_renders_escaped`. |
| 4 | P2 | `untracked_content` | TOCTOU between the metadata check and `File::open` (which follows symlinks). | In-process swap of a regular file for a symlink after the check read `TOCTOU-SECRET\n`; `O_NOFOLLOW` would have failed. | **FIXED structurally** by the same change as #1: there is no longer a check-then-open pair — the object is classified from the handle that was actually opened. |
| 5 | P3 | `untracked_content` | Any path containing `"` was refused, so a worker could hide a content head behind a quote in a filename (porcelain `-z` never quotes). | `?? foo"bar.txt` → `<path not readable>`. | **FIXED**: only the C-quoted form (always a leading quote) is refused; an interior quote is read. Test: `an_interior_quote_is_read_but_a_quoted_porcelain_path_is_refused`. |
| 6 | P3 | `controller.rs` `run_diff` | A path **equal** to the reservation base passes `starts_with`. | Predicate allowed the base itself; prefix confusion (`/base` vs `/basement`, `worktreesEVIL`) was correctly refused component-wise. | **ACCEPTED, unchanged**: the base itself is not a work tree, so `observe_status` refuses it and `run_diff` returns a typed setup error — no observation is widened. Prefix semantics are now pinned by a test (`worktreesEVIL`). |
| 7 | P3 | tests | Coverage gaps; explicitly **no** vacuous tests (the reviewer checked each new test would fail against a wrong implementation). | `--stat` shape unasserted; hostile bytes never placed in the untracked path/content fields; no 3-/4-byte UTF-8 case; no prefix-confusion case on the real `run_diff`. | **FIXED**: `--stat` shape asserted (no `@@`, note empty), hostile untracked path + content pinned inert, 3-byte boundary cut test added, prefix-confusion case added to the controller test. The 4-byte cut is covered by the reviewer's own probe (`utf8_4byte_cut_off1..4` PASS) rather than a repo test: it exercises the same branch as the 2- and 3-byte tests. |

### Reviewer's verification of the claims that held

- `cargo test -p symbiote-repo -p symbiote-desktop --locked` → **exit 0**; all new tests green (it used the system cargo; the rustup proxy is broken under this AppImage).
- Probe results confirming the intended behavior: 2/3/4-byte boundary cuts keep the complete character prefix; genuinely invalid bytes still yield `<non-utf8 content>`; exactly-32 / 35-file caps → `heads=32 omitted=3`; exactly-16 KiB not truncated / +1 byte truncated; untracked directory and symlink not read; `--stat` degradation only for oversized tracked diffs; HTML escaping inert for every payload it tried; containment refused outside/`..`/nonexistent/symlink-escape/prefix-confusion.
- MSRV: no let-chains, no post-1.85 std APIs, no `unsafe` in these crates.

### Re-verification at the fixed head `f09e347a`

```
cargo fmt --all --check                                         -> 0
cargo clippy --workspace --all-targets --locked -- -D warnings   -> 0
cargo test --workspace --locked                                  -> 0  (458 passed / 0 failed)
```

### Not claimed

- The race-winning two-thread TOCTOU is not deterministically testable; the window is closed by construction (open with `O_NOFOLLOW`, classify the opened handle), and the two deterministic escapes (`absolute path`, `symlinked parent`) are pinned by tests.
- The reviewer's `--stat`-hang observation and hardlink probes were left open (this host did not block on `OutputTooLarge`; `protected_hardlinks=1` blocked the cross-filesystem hardlink).
- **No live GUI verification of this increment**, and **live model turns remain gated**: one live native OpenAI Responses API turn and one live Codex 0.118.0 App Server turn require explicit user authorization for credentials and billing. Nothing was spent; no live turn was attempted; the labeled fixture transports remain the only execution path.

PR #569, opened 2026-09-15
feat(stamp): let the wire file own the record's framing and every refusal (#470)

RepairYourTech:
## What this change is

`crates/symbiote-source-stamp/src/wire.txt` becomes the single owner of the record format and of every refusal text both readers print: framing and `encoding`, the rule that reads a record line and the order it reads it in, the marks with their remedies and their precedence, and the malformed, undecodable and unframed texts. The crate embeds the file (`wire.rs`, `include_str!`) and `planning/integrity/source_record.py` reads it, so neither keeps its own copy of a value or a remedy.

Framing is now airtight in both directions. The `end` that ends a record is the one that follows the LF its last line ends with, so a marker the record's own bytes spell does not end it, and a reader that finds one in the bytes it framed refuses the record through the file's `unframed` text instead of reading it short. On the writer's side, `build_stamp` refuses to stamp a locator a record cannot hold — a line feed, a carriage return, or either marker — naming the path and what it holds, so a workspace that builds produces bytes its own readers accept.

The same change splits the stamp crate by concern: `wire`, `walk` (`roots`, `invocation`, `asking`, `paths`), `scan`, `manifest`, and the entry-point proofs beside them.

## Why

Two readers that decide the format separately describe one artifact two ways, and measured, they did. A record whose locator spelled the end marker was framed at that inner marker by both readers, read missing every line after it, and reported a binary as matching a tree it was not built from — a silent wrong answer in the one guard that exists to catch a stale binary. A record whose bytes were not the encoding the file states was raised over by the checker while the crate reported a binary carrying no record. A CRLF record was read as inputs whose locators each ended in a CR, so a line the compiler never read was compared against the tree. A blank line was passed over by both. And a build stamped a locator holding a line feed, then both readers refused its own artifact in a remedy asking for a rebuild that reproduces the same bytes.

## Review findings, addressed in `b3b7fbc3`

* **Target directories are excluded at the package root alone.** `package_sources` skipped `tests`, `examples` and `benches` at every depth, but cargo looks for those target directories at the package root alone. Measured on a probe crate, `src/tests/mod.rs` declared as `mod tests;` — no `#[cfg(test)]` anywhere — is compiled into the binary, so the walk left a compiled file out of the record and a change to it was reported clean. Those three names are now excluded at the root; build output, dependency cache and version-control state stay excluded wherever they sit. The checker's `excluded` keeps the same split, so it can no longer excuse a record the crate would write short.
* **`+toolchain` is one argument.** `subcommand_index` read it as an option that takes a value, so `cargo +nightly b --offline` — an alias — lost its subcommand: no alias was expanded, the line named no manifest, and the record named the run directory's workspace while the fork the build compiled went unnamed and a change to it was reported clean. Measured, the directive is rustup's and is consumed before the cargo this reading reads runs (`cargo +stable build` leaves a build-script parent whose argv is `[<toolchain>/bin/cargo, build]`, and that toolchain's own cargo refuses the directive outright: `error: no such command: '+stable'`). The rule is now stated in the code and pinned by tests rather than left to be read off a skip count.
* **Every refusal states its own sentence.** The shared `malformed` and `undecodable` remedies ended without a period while each reader appended its own sentence, so both printed run-on refusals (`check that build The line is …`). The wire file now ends the sentences it states; a reader adds only its own part.
* **The malformed-over-mark precedence is written down.** `check` reported the malformed line before the mark and the crate stopped at the line as it read, and neither reader recorded that rule. The wire file states it, and the crate's test and the checker's test pin one record carrying both.

## What was verified

* `cargo test --workspace --locked` is **650 passed / 0 failed over 82 suites** after the review fixes, with `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings` and `git diff --check` all clean.
* `planning/integrity` **89** (the new precedence test included), `planning/housekeeping` **11**, `planning` legacy importers **1**.
* The CI record check, replayed **cold on freshly built binaries**: `cargo build --locked -p symbiote-host --bin symbioted -p symbiote-sandbox --bin symbiote-sandbox-launch` then `source_record.py` → exit 0, `symbioted 113 inputs named, 0 unnamed, 19 inputs cargo read, 0 unnamed`, `symbiote-sandbox-launch 40 / 0, 7 / 0`. Every locator in those two records is relative, so a fresh checkout resolves them.
* The `+toolchain` measurement above, and the probe-crate measurement behind the target-directory split, were run in scratch directories with the local toolchain and are reproduced in the code comments beside the rules they settled.
* The commit message and the pull-request title both pass `planning/integrity/closing_keywords.py`, and CI agrees (`closing-keywords` green on `b3b7fbc3`).

A bounded portion of #470. Nothing in this pull request is meant to close an issue, and the rest of #470 stays open.


RepairYourTech:
All four findings in the CodeRabbit review are addressed in `b3b7fbc3`, with the measurements behind them recorded in the code comments beside the rules:

1. `package_sources` now excludes `tests`, `examples` and `benches` at the package root alone (build output, cache and VCS state stay excluded at every depth); the checker's `excluded` keeps the same split, and a walk test pins both halves of it.
2. `subcommand_index` reads `+toolchain` as one argument, with the argv and the refusal a real cargo gives it recorded in the doc comment, and unit tests for both the index and the alias expansion a `+toolchain` line reaches.
3. The shared `malformed` and `undecodable` remedies now end their own sentences, so neither reader prints `check that build The line is …` any more.
4. The wire file states the malformed-over-mark precedence both readers already kept; the crate test and the checker test pin the same record carrying both.

Verification on this head: 650 passed / 0 failed over 82 Rust suites, 89 `planning/integrity`, 11 `planning/housekeeping`, fmt/clippy/`git diff --check` clean, and `source-records` green on freshly built binaries (symbioted 113 named / 0 unnamed).

RepairYourTech:
@coderabbitai review

RepairYourTech:
## Adversarial review of head `d0207074` — no defect found in the diff

Keyed to `d0207074c908db2b791ed192720c12a7cef9fc73` (ten workflow checks green, `mergeStateStatus` CLEAN). I re-derived both changes from the code paths rather than from the diff, and drove real builds instead of trusting the unit tests. Nothing in the diff required a fix, so the head is unchanged since that SHA.

### Drove end to end

**A real crate that stamps a record** (`build.rs` → `build_stamp`, `src/tests/mod.rs` compiled by `mod tests;`, plus an uncompiled `src/tests/data.json`):

* the record names `src/tests/mod.rs` (compiled, previously dropped) and `src/tests/data.json` (not compiled — over-recording, which is what the walk already does for uncompiled assets inside a package);
* changing the module makes `changed_sources` name `/…/src/tests/mod.rs`, so the file the fix was about is actually protected now.

**The forged short record the old rule would have written.** I took the built binary, dropped the single `…\tsrc/tests/mod.rs` line from the embedded record bytes, and ran both rules over it:

| rule | result |
| --- | --- |
| new (target names at the package root alone) | `PROBLEM: probe-short: the record does not name /tmp/stamp-nested-e2e/src/tests/mod.rs, which target/debug/deps/probe-01f5bedd98efba41.d says was read` — exit 1 |
| old (target names at every depth) | `0 unnamed`, `problems: []` — passed |

That is the false green reproduced on **both** readers and then caught by the new rule: the crate-side fix is load-bearing, and `planning/integrity/test_source_record.py::test_a_target_name_below_the_package_root_is_not_an_exclusion` fails under the restored old rule and nothing else does (verified by monkeypatching `EXCLUDED_ANYWHERE`/`EXCLUDED_TARGETS` back).

### What held under the adversarial pass

* **`at_root` cannot misfire.** Every nested path is built by `entry.path()` from the root's own spelling, so a directory entry can never equal the root; the comparison needs no canonicalization, so a root spelled with `..` or a trailing slash behaves identically.
* **The split fails in the safe direction.** Root-only exclusion can only drop a compiled file (false green); including nested ones can only record an extra (rebuild noise). `check()` reports unnamed inputs, content drift and missing/renamed configuration and never surplus names, so the widening cannot turn CI red.
* **The root-level exclusion is not the same hole.** The scan inserts every followed input into `sources` regardless of the exclusion list, so a `#[path = "tests/…"]` or `include!` target is still recorded when the blind walk skips its directory.
* **Both consumers of the shared texts.** The added period cannot double up or run on: the only concatenation sites are `lib.rs` and `source_record.py:820` (both `{remedy} The line is …`), and `undecodable` is rendered whole by `Wire::undecodable` and `undecodable_text`. No other site concatenates these values.
* **`+toolchain` in every argv shape** (plain, toolchain-first, toolchain with options, `--color`/`--config` with values and in `=` form, `--`, no subcommand): the new reading is correct in all of them, and no shape that read correctly before reads wrongly now.
* **Boundaries and hygiene.** `relative.parts[0]` is guarded against an empty tuple; a *file* named `tests` at the package root is still skipped by both readers; overlapping closure packages keep any-true semantics; no stale reference to the removed `EXCLUDED_DIRECTORIES` remains; the new `#![cfg(target_os = "linux")]` test module compiles on the pinned 1.85 toolchain (the MSRV job is green at this head).

### Reported, deliberately not fixed here — one pre-existing silent false green

`dist`, `node_modules`, `target` and `.git` are still excluded by name **at any depth**, and the review's split put `dist` there deliberately. Measured on a second probe crate with `mod dist;` and `src/dist/mod.rs`: the compiled file is **missing from the record** and the checker reports `0 unnamed` / `OK` — the same silent false green this PR fixes for `tests`, still live for a name that is not a cargo target directory at all.

Closing it is a change to what the constant *means* (is generated frontend output under a nested `dist/` recordable, and should it refuse a shipped binary?) rather than to the rule this PR states, so it is left for a separate change. The measured candidate is moving `dist` out of the any-depth set, keeping `target`/`node_modules`/`.git` there.

Two smaller observations, also untouched: `lib.rs`'s summary ("everything under the package that is not a test, example or bench target") is now an approximation only for a target explicitly declared *inside* a nested target-named path — noise, never false green; and the wire file pins the text a malformed line refuses with but never says *which* line when a record holds several, though both readers name the first (pre-existing).

PR #571, opened 2026-09-15
fix(stamp): name every file the compiler finds, and measure before reporting green

RepairYourTech:
## What this change is

Two ways the record read clean while the tree and the binary disagreed, both measured on real builds. They are one failure: a green the guard had not earned. One is on the crate's side (a skipped directory hid a compiled module), one on the checker's (it reported OK having measured nothing).

## 1. A skipped directory hid a compiled module

The walk skips the directory names the wire file's `exclusion` lines list. The compiler does not: it finds a module source by the module's own name, so `src/target/mod.rs`, `src/node_modules/mod.rs` and a package-root `tests/mod.rs` are compiled by `mod target;`, `mod node_modules;` and a `build.rs` that declares `mod tests;` — whatever the wire file says about those names.

Measured on a probe crate whose build stamps a record, with the module compiled (proven by breaking it: the build fails inside it):

| | before | after |
| --- | --- | --- |
| record names `src/target/mod.rs` | no (24 lines, 0 under `src/target`) | yes (26 lines, 1) |
| `source_record.py` | OK | OK |
| `changed_sources` after the module changes, nothing rebuilt | **CLEAN** | **refuses, naming the file** |

`src/node_modules/mod.rs` and the package-root `tests/mod.rs` behave identically — all three were CLEAN before and are refused now.

The fix follows the declaration rather than widening the walk: the scan resolves every plain `mod name;` to `name.rs` and `name/mod.rs` in each directory the declaring file could have been resolved against, tracked through inline-module blocks, and leaves a declaration alone where a `#[path]` names its file. This is why the previous two slices could close `dist` and the nested target names by *walking* them while `target` itself must stay skipped at a package root — what is there is generated, so the answer is to follow the compiler rather than to record the build's own droppings. A skipped directory now bounds the walk and not the build, and the wire file says so where it lists the names.

**No name is added to a record in a tree with no skipped directory.** Checked rather than argued: the locator sets of `symbioted` and `symbiote-sandbox-launch` are identical before and after (115 and 40 names, no additions, no removals).

## 2. The checker reported OK having measured nothing

Completeness is compared against the dep-info of the closure's units, and the check did not require any of them to be there. Measured, against a target directory holding no dep-info at all — what `cargo clean` leaves — it printed `0 workspace inputs read by 0 units` and then `OK`, and a dependency's units alone passed the same way. It now requires a dep-info for at least one unit of the driven package itself and refuses without one, so `OK` cannot be printed by a comparison that never looked at the build.

## Tests that bite

- `scan::tests::a_module_declaration_names_a_file_the_walk_skips` — restoring the walk's skip fails this and nothing else (80 passed, 1 failed).
- `scan::tests::a_plain_module_resolves_where_the_compiler_looks_for_it`, `a_declaration_names_a_module_unless_an_attribute_names_its_file` — the resolution and the `#[path]` exclusion.
- `test_an_oracle_with_no_unit_of_the_driven_package_measures_nothing`, `test_a_dependencys_units_do_not_stand_in_for_the_driven_packages_own` — dropping the own-unit requirement fails exactly these two (92 passed, 2 failed).

## Measured, and needing no change

`include_str!`/`include_bytes!` of a file under an excluded name, a `#[path]` module under `.git`, and a non-Rust file a proc-macro reads while expanding inside a closure package are each recorded and each refused when changed. Declared reads are covered wherever they live, which is what makes the declaration the right seam to close this at.

## Set aside, reported, not changed

- A build script's runtime `fs::read` of a file outside every closure package is outside both readers: the record cannot name what no walk and no declaration reaches, and `changed_sources` could not see it either. Reproduced end to end (stale binary carrying the old value, tree holding the new one, guard CLEAN). Closing it would mean reading cargo's private fingerprint state, and it would still not be visible to the proof side — this repository has no build script that reads outside its package.
- The dep-info oracle identifies a unit by its crate name alone, so a workspace with two same-named targets attributes the other package's unit to the driven binary. Reproduced on a probe (a false `FAIL`, not a silence). This workspace has no such collision: 27 targets across 23 local packages, all distinct.

## Verification

- `cargo test --workspace --locked`: 655 passed, 0 failed, on stable and on the pinned 1.85.0.
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- `planning/integrity` 94, `planning/housekeeping` 11, `planning/test_legacy_importers.py` OK.
- Record check green on binaries freshly built from this head: `symbioted` 113 named / 0 unnamed / 19 cargo inputs / 0 unnamed; `symbiote-sandbox-launch` 40 / 0 / 7 / 0.


RepairYourTech:
Measurements this change rests on, for the record, keyed to `582d90c468645516b74afaf7800336d82c0724f8`.

**The reproductions, on a real crate whose build stamps a record** (`cargo build` on `probe`, whose guard binary is the package's own, built and run under cargo so the record's pinned `env:CARGO` matches):

```
# before this change
mod target;        -> record: 24 lines; 0 matching 'src/target'
                      checker: OK: every workspace input the compiler read ... is named
                      guard after the module changed, nothing rebuilt: CLEAN: the record matches the tree
mod node_modules;  -> record: 24 lines; 0 matching 'src/node_modules'; guard: CLEAN
build.rs 'mod tests;' with tests/mod.rs -> record names nothing there; guard: CLEAN

# after
mod target;        -> record: 26 lines; 1 matching 'src/target'
                      guard:   /tmp/stamp-escape/src/target/mod.rs
mod node_modules;  -> record: 26 lines; 1 matching 'src/node_modules'
                      guard:   /tmp/stamp-escape/src/node_modules/mod.rs
tests/mod.rs       -> guard: REFUSED: 1 input(s) differ   /tmp/stamp-escape/tests/mod.rs
```

The module really is compiled: replacing `src/target/mod.rs` with `this is not rust` fails the build inside it (`error: expected one of \`!\` or \`::\`, found \`is\``).

**The checker's zero-evidence green**, against a target directory holding no dep-info:

```
before: probe: 0 workspace inputs read by 0 units, 24 named by the record, 0 unnamed; 5 inputs cargo read, 0 unnamed
        OK: every workspace input the compiler read, and every input cargo read to build it, is named by the record that drove it
        (with only a dependency's dep-info present: 2 workspace inputs read by 13 units ... OK)
after:  FAIL: the target directory holds no dep-info for any unit of the driven package
```

**The classes that came back sound**, each probed the same way and each refused after a change: `include_str!` of a file under an excluded name (record names it), `#[path]` module under `.git` (record names it), a non-Rust file a proc-macro reads while expanding inside a closure package (record names it).

**No name added to a real record**: the locator sets of `symbioted` (115) and `symbiote-sandbox-launch` (40) are identical before and after, by name.

**Left for a separate change**, both reproduced and neither a silence: a build script's runtime `fs::read` of a file outside every closure package (needs cargo's private fingerprint state to see, and the proof side still could not), and the dep-info oracle's crate-name-only unit identity, which turns a workspace with two same-named targets into a needless refusal (this workspace has none: 27 targets, 23 packages, all distinct).

Local at this head: 655 workspace tests / 0 failed on stable and pinned 1.85.0, fmt and strict clippy clean, `planning/integrity` 94, `planning/housekeeping` 11, legacy importer suite OK, and the record check green on freshly built binaries.

PR #572, opened 2026-09-15
fix(integrity): identify a build unit by cargo fingerprint, not by its crate name

RepairYourTech:
## What this change is

The checker identified a build unit by the crate name in its dep-info file name. A crate name is shared by every package carrying a target of that name, so on a workspace whose targets collide the wrong package's unit was attributed to the driven binary — and that is the seam the previous slice's rule ("a green rests on a unit of the driven package's own build") leans on.

## Reproduced on a two-package workspace

`probe` (whose binaries are `probe` and `guard`) and a second package whose binary is also `guard`, both built into one target directory. Cargo wrote `deps/guard-cf49597cb80e0847.d` (from `probe/src/bin/guard.rs`) and `deps/guard-25d04fabbb313b6d.d` (from the other package's `src/main.rs`) — differing in nothing but the hash.

**A false green.** With `probe`'s own units gone — what `cargo clean -p probe` leaves — the *other* package's file satisfied the own-unit rule:

```
before:  probe: 2 workspace inputs read by 13 units, 23 named by the record, 0 unnamed; 5 inputs cargo read, 0 unnamed
         OK: every workspace input the compiler read ... is named by the record that drove it
control: deleting every dep-info instead -> FAIL: ... holds no dep-info for any unit of probe
after:   FAIL: ... holds no dep-info for any unit of probe
```

The control is the proof that the foreign unit was the only thing standing in for the missing build.

**A false attribution.** With both present, the foreign unit was counted as one of `probe`'s closure — **15 units where the build has 14** — and its reads were compared against the record. On a workspace whose member shares the name it read as a needless refusal (`the record does not name …/guard/src/main.rs, which …/deps/guard-<hash>.d says was read`). After: 14 units, no problem, and that workspace is clean.

## The identity, from cargo's own output

`<profile>/.fingerprint/<package>-<hash>/` — one directory per unit, named after the package that compiles it **and the unit's metadata hash, which is the hash the unit's dep-info file carries**. Measured in the collision workspace: `guard-25d04fabbb313b6d.d` → `other-25d04fabbb313b6d`, `guard-cf49597cb80e0847.d` → `probe-cf49597cb80e0847`. Inside the directory, the per-target JSON names the target (`bin-guard.json`), which is what confirmed the ownership above.

So a unit is taken only where cargo's fingerprint places its hash in a package that is in the build's closure **and** that package has a library, binary or proc-macro target of that crate name; a hash no fingerprint places, or one more than one places, is not an identity and is not counted at all. Build-script dep-info is unchanged — it is taken by directory, which already places it in the package it is named for.

Coverage in this tree: **2775 of 2776** dep-info files are placed by a fingerprint; the one that is not carries no hash in its name at all (`symbiote_desktop_lib.d`) and is not a unit of either driven package.

## Tests that bite

- Reverting the *selection* to crate-name matching fails `test_a_foreign_unit_is_not_compared_as_one_of_the_driven_binaries` alone (95 passed, 1 failed).
- Reverting the *own-unit rule* to crate-name matching fails `test_a_unit_of_another_package_is_not_one_of_the_driven_packages_own` alone (95 passed, 1 failed).

Both are needed and neither substitutes for the other: the selection fix removes a unit of a package outside the closure, and the own-unit rule refuses a *closure* dependency that merely shares the target name. The fixtures now write the fingerprint a real cargo writes, since that is where the identity comes from.

## Both readers still agree, and the records do not move

No Rust file is touched: the crate writes exactly the same records. Verified rather than assumed — the locator sets extracted from `symbioted` and `symbiote-sandbox-launch` are identical to those captured at the last merge, and the checker's own numbers are unchanged: `symbioted` 870 units read / 19 cargo inputs / 113 named / 0 unnamed; `symbiote-sandbox-launch` 542 / 7 / 40 / 0.

## Verification

- Python: `planning/integrity` 96, `planning/housekeeping` 11, `planning/test_legacy_importers.py` OK.
- `cargo test --workspace --locked`: 655 passed, 0 failed on stable and on the pinned 1.85.0.
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- Record check green on binaries built from this tree.

## Scope

This is the identity defect and its tests. The build-script read outside every closure package stays a reported, unfixed limitation, and no new capability is added beyond making the existing evidence trustworthy.


RepairYourTech:
Reproduction and identity evidence, keyed to `a8732a07c62ec50927dbb584eb7b4a28040815bc`.

**The collision, built for real** (`/tmp/stamp-collision`): package `probe` with binaries `probe` and `guard`, plus a second package whose binary is also `guard`, both into one target directory. The two dep-info files:

```
guard-25d04fabbb313b6d.d   -> fingerprint other-25d04fabbb313b6d  (the other package's src/main.rs)
guard-cf49597cb80e0847.d   -> fingerprint probe-cf49597cb80e0847  (probe's src/bin/guard.rs)
probe-a19ffa0c08b57a0c.d   -> fingerprint probe-a19ffa0c08b57a0c
```

**False green** — `probe`'s own units removed, the foreign one left (the state `cargo clean -p probe` leaves):

```
before:  probe: 2 workspace inputs read by 13 units, 23 named by the record, 0 unnamed; 5 inputs cargo read, 0 unnamed
         OK: every workspace input the compiler read, and every input cargo read to build it, is named by the record that drove it
control: every dep-info removed -> PROBLEM: ... holds no dep-info for any unit of probe  (+ FAIL)
after:   PROBLEM: ... holds no dep-info for any unit of probe  (+ FAIL)
```

**False attribution** — both units present:

```
before:  probe: 3 workspace inputs read by 15 units, 23 named by the record, 0 unnamed   (the build has 14)
after:   probe: 3 workspace inputs read by 14 units, 23 named by the record, 0 unnamed
```

and on a workspace whose *member* shares the binary name (the earlier probe at `/tmp/stamp-codegen`), before: `PROBLEM: probe: the record does not name …/guard/src/main.rs, which …/deps/guard-9714db9933b1d237.d says was read`; after: `OK`, no problem.

**The repo's own measurement is unchanged**: `symbioted` 870 units / 19 cargo inputs / 113 named / 0 unnamed; `symbiote-sandbox-launch` 542 / 7 / 40 / 0 — and the locator sets extracted from both binaries are identical to those captured at the last merge.

**Coverage of the identity** in this tree: 2775 of 2776 `deps/*.d` are placed by a fingerprint; the one that is not (`symbiote_desktop_lib.d`) has no hash in its name at all and is not a unit of either driven package. A unit no fingerprint places is not counted, and the own-unit rule is what then refuses the comparison — which is exactly why that rule and the identity fix are both needed.

**Bite checks at this head**: selection by crate name again → `test_a_foreign_unit_is_not_compared_as_one_of_the_driven_binaries` fails alone; own-unit rule by crate name again → `test_a_unit_of_another_package_is_not_one_of_the_driven_packages_own` fails alone.

PR #574, opened 2026-09-15
fix(integrity): excuse the directory the check was given, not a name a walk skips

RepairYourTech:
## What this closes

The completeness rule excused a read when its **path carried a name** the wire file's `exclusion` lines skip. That one rule was wrong in both directions at once, and both faces are reproduced below.

### Face A — a legitimate build refused (false red)

The generated directory was hardcoded to the name `target`. A build driven into `<workspace>/build-output` is the same build, but the rule reading the name `target` demanded every dep-info file cargo writes under `build-output/debug` — files no record can name, because cargo wrote them.

Measured on a fresh clone of `main`:

```
$ CARGO_TARGET_DIR=$PWD/build-output cargo build --locked -p symbiote-host --bin symbioted \
    -p symbiote-sandbox --bin symbiote-sandbox-launch
$ python3 planning/integrity/source_record.py --binary symbiote-host:symbioted \
    --binary symbiote-sandbox:symbiote-sandbox-launch
symbioted: 215 unnamed
symbiote-sandbox-launch: 138 unnamed
353 problems
```

The same build with the directory *outside* the workspace passed, so the name was the whole of the cause.

### Face B — a file the build read excused (false green)

`build.rs`'s `mod tests;` puts `tests/mod.rs` in the record — the crate duly names it — but the checker excused it by name, so a **forged short record** passed:

```
$ python3 planning/integrity/source_record.py --binary probe:probe     # record line removed
OK: 0 unnamed
$ target/debug/guard          # the crate's own reader, same binary bytes
CLEAN
```

Control: with the excusal disabled and nothing else changed, the same forged bytes are **refused**, naming `tests/mod.rs` as read by the build-script unit — which shows the excusal was what hid it.

## The change

The excusal is **positional and nothing else**:

* A read is excused only when it lies under the build's own output directory — the directory the check was **given** (cargo's `target/<profile>`, or whatever a build was driven into), because cargo wrote it and the record names the build script that did.
* Every other input a unit read must be named, a `tests`, `examples`, `benches`, `node_modules` or `.git` path among them. The file's own rule is that a name is skipped only where skipping it cannot hide a file the build read; a check is the reader that can **verify** that claim rather than mirror it.
* The words `target/` left the checker: `excluded()`, `EXCLUDED_AT_ROOT`, `EXCLUDED_ANYWHERE` and `GENERATED_DIRECTORY` are gone. `EXCLUSIONS` is still parsed because the file is one copy that must be accepted whole — a scope this tool did not know is a file the crate could skip differently.
* `crates/symbiote-source-stamp/src/wire.txt` states the rule where the format lives, with both measurements; `planning/integrity/README.md` no longer keeps a third copy of which names are skipped and where.
* Nothing in the crate moves: the walk, what it skips, and what the record names are unchanged, so the two readers still agree on what a record must name.

**Why this is safe to narrow:** the name-based excusal fired **zero** times in this workspace *and* in an `--all-targets` clone, so no state this repo builds depended on it.

## Verification

* Face A: `353 problems` → `OK`, on the same clone and the same build.
* Face B: forged short record `OK`/`CLEAN` → refused (`holds no dep-info…`/`does not name tests/mod.rs`), and back to `OK` when the line is restored.
* Each face has a test that fails **alone** when it is reverted: restoring the name-based excusal fails `test_a_read_under_a_name_a_walk_skips_is_still_the_records_business`; reading the name `target` again fails the two generated-directory tests.
* Python suites green (integrity 100, housekeeping 11, legacy importers).
* `cargo test --workspace --locked`: 655 passed / 0 failed on stable **and** the pinned 1.85.0; `cargo fmt --all --check` and strict clippy clean.
* Record check green on binaries freshly built by each toolchain, unchanged numbers: `symbioted` **113 named / 0 unnamed**, `symbiote-sandbox-launch` **40 / 0**.
* Other legitimate states re-measured and still passing: a fresh clone built once with CI's command, `--release`, a copied artifact (uplift relation gone, target fallback applies), `cargo clean -p` then rebuild, `--all-targets`, a target directory outside the workspace, and two toolchains' artifacts in one target directory.

## Out of scope

The build-script read outside every closure package stays a reported, unfixed limitation — it is a missing capability rather than a rule that excuses evidence.


RepairYourTech:
## Review result, keyed to head `43a38d7e788a45ebb4860b53a634c1006362c7e1`

Both faces of the excusal rule were driven end to end on real builds, before and after.

**Face A — 353 problems → OK.** Fresh clone of `main`, built with `CARGO_TARGET_DIR=$PWD/build-output`: the pre-slice checker refused with **353 problems** (215 unnamed for `symbioted`, 138 for `symbiote-sandbox-launch`) because it excused reads by the *name* `target`. With `--target-dir` honoured positionally, the same build is `OK` with both binaries at 113 named / 0 unnamed and 40 / 0. Bounds measured: the same build with the output directory *outside* the workspace passed before the slice too, so the name was the entire cause.

**Face B — clean → refused, with a control.** A probe package whose `build.rs` declares `mod tests;` puts `tests/mod.rs` in the record; deleting that one line from the binary's record bytes left the pre-slice checker reporting `OK, 0 unnamed` **and the crate's own guard reporting `CLEAN`**. Control: disabling the name-based excusal and changing nothing else refused the same bytes, naming `tests/mod.rs` as read by the build-script unit — so the excusal, not the evidence, was what hid it. Restoring the line returns both readers to clean.

**What holds.** The name-based excusal fired **zero** times in this workspace and in an `--all-targets` clone, so nothing this repo builds depended on it. Every other legitimate state still passes: fresh clone built once with CI's command, `--release`, a copied artifact (uplift relation gone, target fallback applies), `cargo clean -p` then rebuild, `--all-targets`, an output directory outside the workspace, and two toolchains' artifacts in one output directory.

**Each face is pinned by a test that fails alone when reverted** — restoring the name-based excusal fails `test_a_read_under_a_name_a_walk_skips_is_still_the_records_business`; reading the name `target` again fails the two generated-directory tests. The crate is untouched, so both readers still agree on what a record must name, and the rule now lives once, in `crates/symbiote-source-stamp/src/wire.txt` (the README's third copy of which names are skipped, and where, is gone). CI's `source-records` job — which runs the changed checker against binaries CI itself built — is green, so the narrowing produced no false red on the real pipeline. All ten checks green on `43a38d7e788a45ebb4860b53a634c1006362c7e1`; Python suites 100 / 11 / legacy; 655 Rust tests / 0 failed on stable and pinned 1.85.0; fmt and strict clippy clean.

**Pre-existing, deliberately not in this change:** the build-script read outside every closure package stays a reported, unfixed limitation — a missing capability rather than a rule that excuses evidence. The crate's nested-`dist` walk semantics and the walk itself are unchanged.

PR #575, opened 2026-09-15
test(stamp): keep the guard's reproduced shapes as real-build tests

RepairYourTech:
## Why

Six slices of this work were proved the same way: a scratch workspace built by real cargo, the record read out of the binary's own bytes, both readers put to what the build produced. Every one of those scratch workspaces was then thrown away. What held the fixes afterwards was a unit test on a synthetic tree plus a paragraph of prose, so a regression in what a record covers would have survived until another audit found it.

This makes the reproductions that matter permanent, following the pattern `crates/symbiote-source-stamp/tests/entry_point.rs` already sets: a fixture workspace built by the cargo that is running the test, every invocation `--offline`, a fresh directory per case, no fixture outside it, and the repository's own checker run on the same bytes as this crate's reader. Nothing new is claimed by the fixture; it is the audit's own arrangements, in the suite.

```
crates/symbiote-source-stamp/tests/record_guard.rs   4 tests, ~4s
planning/integrity/README.md                          no longer says these survive only in an audit
```

## The four shapes, and what each fails on

**1. A module the compiler finds under a name the walk skips is named by the record.** `mod target;` → `src/target/mod.rs` (a name the wire file skips at every depth) and a build script's `mod tests;` → `tests/mod.rs` (a name it skips at a package root alone). The test asserts on the record a real build wrote, requires the checker to accept it, and then changes the skipped module to prove the record **covers** it rather than mentioning it.

**2. A record with one line removed is refused.** The line naming `tests/mod.rs` is dropped from the binary's bytes — the shape a build would have written had it not read that file, and the one shape no reading of the record alone can tell from a build's own. The checker must refuse and name the file; the crate's reader accepts the same bytes, which is its documented division of labour (it compares the content of the files a record names, not whether a name is missing), asserted so that the pair's roles cannot drift silently.

**3. A build driven into `<workspace>/build-output` is complete.** The record cannot name the files cargo wrote under that directory; the check excuses them by the directory it was *given*. The test also asserts the record names nothing there, so an acceptance cannot be the record happening to name cargo's own output.

**4. Neither another package's unit nor another unit of this package is evidence.** Three units here carry the crate name `probe`: the driven binary, a library of the same package, and a library of a package the driven binary's build never reaches.

## Proof each shape bites

Each was checked by installing the reader it pins and confirming the failure:

| test | installed reader | measured failure |
|---|---|---|
| 1 | crate at `d7de5e47~1` (before declarations were followed) | the record stops naming `src/target/mod.rs` and `tests/mod.rs`; all four tests fail |
| 2, 3 | checker at `b05315a1~1` (excusing names) | 2: the forged line is excused and the record reported clean. 3: 37 problems, one per file cargo wrote under `build-output` (353 on this workspace) |
| 4, face a | checker at `635097df~1` (identity by crate name) | the check refuses the binary over `ws/other/src/lib.rs`, another package's source |
| 4, face b | checker at `b05315a1~1` (any unit of the package) | the package's own library satisfies the rule and the record is called complete having never been compared with this build |

Face b is the array that makes the point: the pre-slice checker printed OK.

## What this harness does not cover, and why

* **`907097f1`** (the wire file owning framing and refusals) is already pinned by real-build tests in `entry_point.rs` — `both_readers_read_a_record_line_by_the_wires_own_rule` and `a_build_that_can_name_neither_marks_both_and_the_cargo_is_reported` — so it is not restated here.
* **`7011b93c`**'s exclusion list is pinned by the crate's unit tests (`a_directory_of_a_target_name_below_the_root_holds_compiled_files`, `a_directory_no_build_generates_holds_compiled_files_wherever_it_sits`), and its real-build consequence is shape 1. It is *not* independently observable from a completeness check any more: since the scan follows declarations and includes, anything a skipped name could hide is recorded by the declaration that reaches it.
* A read with no oracle at all stays the reported limitation. Probing it inside this fixture: a build script that `fs::read`s `tests/spec.txt` at run time is invisible to cargo's per-unit dep-info, so the checker neither requires it nor refuses it — the walk-vs-build divergence, unchanged and unfixed.

## Verification

`cargo test --workspace --locked`: **659 passed, 0 failed** on stable and on the pinned 1.85.0 (the new file 4 passed on each; the crate's suite is 97 in total, up from 93). `cargo fmt --all --check` and `cargo clippy --workspace --all-targets --locked -- -D warnings` are clean. The three Python suites are green and the record check on freshly built binaries is unchanged: `symbioted` 113 named / 0 unnamed, `symbiote-sandbox-launch` 40 / 0. Runtime added to the suite: ~4s.


RepairYourTech:
## Verification, keyed to head `a760fa0c`

Ten checks green on this SHA, `mergeStateStatus` CLEAN. The one that matters for this change is `contracts`: it runs `cargo test --workspace --locked` on **both** stable and 1.85.0, so the new file's four fixture builds — each invoking the cargo that is running the test, `--offline`, in a fresh temporary directory, and each running `python3 planning/integrity/source_record.py` on the artifact that build produced — passed in CI's own environment rather than only here.

Local, on this SHA: 659 passed / 0 failed on stable and on the pinned 1.85.0 (93 → 97 in the stamp crate), `cargo fmt --all --check` clean, `cargo clippy --workspace --all-targets --locked -- -D warnings` clean, the three Python suites green, and the record check on freshly built binaries unchanged at 113 named / 0 unnamed and 40 / 0.

Each shape was verified to fail when the reader it pins is reverted, by installing that reader and re-running the file: the crate at `d7de5e47~1` (no declaration-following) fails all four — the record stops naming `src/target/mod.rs` and `tests/mod.rs`; the checker at `b05315a1~1` (excusing names) fails the forged-record test (the line is excused, record reported clean) and the custom-directory test (37 problems, 353 on this workspace); the checker at `635097df~1` (identity by crate name) fails the collision test's first face, refusing the binary over `ws/other/src/lib.rs`; and the pre-producing-unit checker fails its second face by accepting the record — the package's own library satisfied the rule.

Not covered, deliberately, and stated in the PR body: `907097f1` is already pinned by the real-build reader-agreement tests in `entry_point.rs`; `7011b93c`'s list is pinned by the crate's walk unit tests and its real-build consequence is shape 1, since the scan now reaches anything a skipped name could hide; and a read cargo's per-unit dep-info cannot see at all — probed inside this fixture with a build script that reads `tests/spec.txt` at run time — remains the reported limitation.

PR #576, opened 2026-09-15
test(stamp): share the real-build scaffolding both guard harnesses need

RepairYourTech:
## Why

Two integration test binaries in this crate now build real cargo fixture workspaces, and the mechanics sat in both: how a fixture directory is made and cleaned, how cargo is invoked, how the record is found under the build directory, how this crate's reader is put to an artifact, and how `source_record.py` is run on the same bytes. Measured, that is **~115 code lines of the same thing in two places** — including the two whose drift would quietly weaken a proof rather than fail one: the cargo invocation and the checker's own command line.

`tests/support/mod.rs` holds those mechanics once, following `crates/symbiote-host/tests/support` and its own reasoning — *"a second copy of the invocation would let one harness prove something the other does not"*. What stays in each file is what makes it that file's harness: the fixture a case writes, the helpers only it needs (`fork`, `artifact`; the unit identity, the forged record, `find`), and every assertion.

## The measurement, because this is a concision change

| | before | after |
|---|---|---|
| `entry_point.rs` | 1178 | 1104 |
| `record_guard.rs` | 544 | 485 |
| `support/mod.rs` | — | 189 |
| total lines | 1722 | 1778 |
| **code lines** (no comments or blanks) | **1230** | **1226** |

So the saving is **ownership, not size**, and I want that stated plainly rather than implied: each call site keeps a thin same-named wrapper, which is exactly what leaves every test body untouched. Calling through `fixture.workspace` directly would have cut ~40 more lines and rewritten the call sites of all sixteen tests — not a trade worth taking for a change whose point is that the mechanics cannot drift.

What is shared: the temporary workspace (created canonical, removed on drop), `cargo`/`build` with `--offline` and `CARGO_TARGET_DIR`, `record` (the build directory, both levels, exactly one), `verdict`/`differences` with the package directory named, `checker` (interpreter, checker path, workspace, profile directory, offline resolution) and the wire-file parsers. What is not: either fixture, or anything either file asserts.

## Verification

* **No test lost coverage and no assertion moved**: each of the sixteen `#[test]` functions was extracted from both revisions and compared — all sixteen byte-identical. Counts unchanged: 81 unit / 12 entry-point / 4 record-guard.
* **Both binaries still drive real builds**: the fixtures are unchanged and `contracts` runs them on CI's own runner.
* **The checker still gets the same bytes**: reinstalling the checker that excuses a read by name still fails the forged-record test through the shared command line, and all four pass again with the current checker.
* `cargo test --workspace --locked`: **659 passed / 0 failed** on stable and on the pinned **1.85.0**. `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean. Python suites green. Record check on freshly built binaries unchanged (113 named / 0 unnamed and 40 / 0).

Production code, the checker's rules and the format in `wire.txt` do not move, and the build-script read outside every closure package stays the reported limitation.


RepairYourTech:
## Verification, keyed to head `c6424790770664e3065f2f58b56feb2c7fb34327`

Ten checks green, `mergeStateStatus` CLEAN, CodeRabbit approved with no actionable comments. The checks that matter for this change: `contracts` on **both** toolchains ran the sixteen real-build tests through the shared scaffolding (5m09s / 5m13s, both pass), and `source-records` is unchanged.

**No test lost coverage and no assertion moved** — not a claim, a check: each of the sixteen `#[test]` functions was extracted from both revisions by parsing the file (the `#[test]` marker to the closing brace at column zero) and compared textually. All sixteen are byte-identical; the counts are unchanged at 81 unit / 12 entry-point / 4 record-guard.

**The checker still receives the same bytes.** Through the shared command line, reinstalling the checker that excuses a read by name still fails the forged-record test (the record reported clean with `tests/mod.rs` missing), and all four pass again with the current checker.

**The measurement, stated plainly.** `entry_point.rs` 1178 → 1104, `record_guard.rs` 544 → 485, `support/mod.rs` 189 new: **1722 → 1778 total lines, 1230 → 1226 code lines**. The duplication removed is real — ~115 code lines of the same cargo invocation, record lookup, both readers' invocation, temp-workspace lifecycle and wire parsing — but the file count does not fall, because every call site keeps a thin same-named wrapper and that is what keeps the test bodies untouched. The win is a single owner for the mechanics, not fewer lines, and the PR body says so rather than implying otherwise.

Local, on this SHA: 659 passed / 0 failed on stable and on the pinned 1.85.0; `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean; the three Python suites green; the record check on freshly built binaries unchanged at 113 named / 0 unnamed and 40 / 0. Production code, the checker's rules and `wire.txt` are untouched, and the build-script read outside every closure package stays the reported limitation.

PR #577, opened 2026-09-15
test(integrity): assert the oracle's cargo facts against a real build

RepairYourTech:
## Why

The fixtures in `test_source_record.py` write cargo's layout by hand — fingerprint directories, dep-info names, the relation between a unit's output and the artifact a binary is — and one of them already wrote it wrong: the fingerprint's extensionless file rather than the `<kind>-<target>.json` cargo writes, so a test pinned a layout the real thing did not have. Nothing in a hand-built fixture can catch that, because the fixture is the thing being trusted.

This adds `planning/integrity/test_source_record_real_cargo.py`: one scratch workspace built by the real cargo (a package whose binary includes a file its build script generates, a library of the same package, and a second package whose library carries the driven binary's crate name), in two target directories — `target` and `build-output`. Three facts, each of which only real output can decide:

| case | what is asserted | measured on the build |
|---|---|---|
| **the producing unit** | the artifact's file is shared with exactly one output in `deps`, and the hash off that relation identifies the unit whose dep-info the check measures | one shared output; **three** dep-info files spelled `probe-<hash>.d` |
| **cargo's fingerprint layout** | the unit's fingerprint is `<package>-<hash>/<kind>-<target>.json` — `probe-<hash>/bin-probe.json` — and it names the target the unit compiles | the shape the hand fixture got wrong |
| **which package a same-named unit belongs to** | the other package's `deps/probe-<hash>.d` is placed by cargo in `other` (fingerprint `other-<hash>/lib-probe.json`) and is **not** counted as evidence for the driven package | a real unit of a package the driven build never reaches |
| **a target directory that is not `target`** | the build into `<workspace>/build-output` leaves real reads under it (the generated file the binary includes among them); giving that directory excused all of them, giving any other directory requires them | 3 reads excused, 8 required without it |

## What this does and does not replace

* It reads **no record**: what is asserted is what cargo wrote. The crate's real-build harness (`tests/record_guard.rs`) drives the checker *end to end* and asserts **verdicts** on a record-bearing binary; it cannot see inside the oracle, so the shapes do not overlap even where the intent does.
* The hand-built fixtures in `test_source_record.py` stay: most shapes a rule has to refuse are not things a build can cheaply be made to produce.
* **Residual, stated rather than implied:** the hand-written fixtures are still hand-written. Nothing automatically compares them to cargo, so a fixture that drifts from real output still needs a real-output test to notice — which is what this module is, for the three facts that matter most.

## Verification

Every case was checked to **bite** by reverting the rule it covers in place and re-running:

| reverted rule | failing case | the failure |
|---|---|---|
| excusing reads by the name `target` | target-directory | the excused set becomes the build's own output instead of the three sources |
| crediting a unit by its crate name | collision | the other package's unit is counted, `package='other'` among the driven package's units |
| returning no producing unit (no uplift) | producing unit | `driven_unit` is `None` where the shared output names a hash |

With the checker restored: the module is **3 passed**, the integrity suite **103** (99 + 3, discovery), housekeeping and the legacy importers unchanged. Runtime added: **under a second** — the scratch workspace has no dependencies, so `--offline` has nothing to fetch (`--locked` does not apply to a generated workspace with no committed lockfile; the build writes the one the checker's own `cargo metadata --locked` then reads). It **skips with the reason** where no cargo can be run, rather than failing.

Nothing in production code, `wire.txt` or the checker's rules moves; the build-script read outside every closure package stays the reported limitation.


RepairYourTech:
## Verification, keyed to head `f8eb8c5ee0d4f05b1b512ce7945720b5b6149c48`

Ten checks green, `mergeStateStatus` CLEAN.

**The real-cargo cases ran in CI, they did not skip.** `offline-validation` reports `Ran 103 tests in 11.226s` and `OK` with **zero** occurrences of \"skipped\", and the three new cases appear by name — `test_source_record_real_cargo.RealCargoOutputTests.test_the_unit_that_produced_an_artifact_is_the_one_cargo_wired_to_it`, `…test_a_same_named_unit_of_another_package_is_not_this_packages_unit`, `…test_a_target_directory_of_its_own_is_excused_where_it_is_given` — so cargo was available on the runner and three scratch workspaces were really built there (11.2s against the 0.4s the suite takes with no build). The job needs no Rust setup step: the runner ships cargo, and the module skips with its reason where one is not to be had.

**Each case bites when the rule it covers is reverted**, checked in place rather than argued: excusing reads by the name `target` makes the target-directory case fail with the build's own output in the excused set; crediting a unit by its crate name makes the collision case fail with the other package's unit counted (`package='other'`); returning no producing unit makes the uplift case fail with `driven_unit` `None`. Restored, the module is 3 passed and the checker's file is unmodified.

**Nothing else moved.** Housekeeping 11 and the legacy importers unchanged; the Rust workspace is untouched, so the `contracts` jobs on both toolchains and `source-records` are unaffected and green. Production code, `wire.txt` and the checker's rules are unmodified.

**What the real output now owns that the hand-built fixtures cannot:** the hash cargo gives a unit (`unit_hash` reads it rather than being written by the fixture), the relation that ties an artifact to the unit that produced it (the profile name and the `deps` output are one file), cargo's fingerprint layout (`<package>-<hash>/<kind>-<target>.json` — the shape the fixture had wrong), which package a same-named unit belongs to, and the fact that a legitimate build leaves real reads under its own target directory. The residual, stated rather than implied: the hand-written fixtures are still hand-written, so a fixture that drifts from real output needs a real-output test to notice — which is what this is, for the three facts that matter most.

PR #578, opened 2026-09-15
test(integrity): skip the real-cargo cases where no cargo can be run

RepairYourTech:
## The defect

The real-cargo cases decide whether to run by taking `CARGO` or a `cargo` on PATH, and **`CARGO` is not proof that the binary it names is there**. Measured, before this change:

```
$ env PATH=/usr/bin:/bin CARGO=/nonexistent python3 planning/integrity/test_source_record_real_cargo.py
FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent'
Ran 0 tests in 0.002s
FAILED (errors=1)
```

`setUpClass` let the `FileNotFoundError` escape `subprocess`, so a state with no cargo to be had reported an **error** where the module's own contract — and the task that added it — says **skip with a stated reason**. The neighbouring case (`CARGO` empty, no cargo on PATH) already skipped, which is what hid it.

## The fix

Ask the candidate for its version before anything is built, and pass over one that cannot answer: a usable `CARGO`, else the PATH's `cargo`, else skip with that reason. One helper, and `setUpClass` asks it instead of trusting the variable.

## Verification, in the four configurations

| state | result |
|---|---|
| normal | 3 passed |
| `CARGO=/nonexistent`, cargo on PATH | 3 passed — the fallback reaches the PATH's cargo |
| no cargo anywhere, `CARGO=/nonexistent` | `skipped 'no cargo that can be run, from CARGO or from PATH, so no real build can be made and these cases would prove nothing'` |
| no cargo anywhere, `CARGO` empty | the same skip |
| the last state under discovery | 100 ran, 1 skipped, `OK` (was a failure) |

The normal suite is unchanged at 103 passed; nothing else in the module, the checker or `wire.txt` moves.


RepairYourTech:
## Verification, keyed to head `c2317e43`

Ten checks green, `mergeStateStatus` CLEAN. `offline-validation` ran the suite as **103 tests with zero skips** and the three real-cargo cases by name, so the fallback made no difference on a runner that has cargo — and the four configurations in the PR body are what the skip path was checked against, including the one that used to be an error.

Nothing else moved: housekeeping 11 and the legacy importers unchanged, the Rust workspace untouched, and no change to the checker's rules, `wire.txt` or any other file.

PR #579, opened 2026-09-15
test(integrity): compare the hand-built fixtures against cargo's own output

RepairYourTech:
## Why

`test_source_record.py` builds cargo's layout by hand — it decides how a unit's dep-info is spelled, where its fingerprint sits, what it is called and which file is the unit's artifact — and nothing in that suite can notice a decision cargo does not make, because the fixture is the thing being trusted. The previous slice closed three facts about *cargo* with a real build; this one closes the other direction, the fixture's own letters.

One drift was already found this way (a fingerprint's extensionless file with no `.json`). This pass found a second, by comparison:

| the fixture's shape | what it wrote | what cargo writes |
|---|---|---|
| a dep-info's inputs | **every input absolute**, `str(root / read)` | a package's own sources **relative to the directory cargo ran in** (`probe/src/main.rs`), cargo's own generated paths absolute |

That is not cosmetic: the relative branch a real build's every source takes was reached by **no test of the hand suite** — measured, dropping the checker's relative resolution failed 0 of its 60 tests then, and 5 now.

## What changed

* `Fixture.unit` writes each read **the way the test names it**, which is the way a build was given it. The one test that declares a read under a build output directory now names it absolutely, which is how cargo passes one.
* `HandFixtureAgreementTests` (in `test_source_record_real_cargo.py`) puts the same shapes in front of a build by the real cargo and asserts the fixture's letters are cargo's — two rows per comparison, asserted to have the same shape rather than described as agreeing:

| shape compared | cargo's side | the fixture's side |
|---|---|---|
| **a dep-info's spelling** | `probe/src/main.rs` relative, the generated input under `out/` absolute | a source relative, a file under its output directory absolute — asserted equal as *relativity* per token |
| **a dep-info's contents** | the `# env-dep:OUT_DIR=` comment, a bare `path:` rule, `hyphen-demo/src/a\ b.rs` with the space escaped | the hand parser's own `DepInfoParsingTests` feed it exactly those three, and the parser recovers the unescaped path from cargo's bytes |
| **a fingerprint** | `<package>-<hash>/bin-<target>` **and** `.json` | the pair `bin-demo` + `bin-demo.json` — the file an earlier fixture left out |
| **identity spellings** | `hyphen_demo-<hash>.d` underscored beside the hyphenated `bin-hyphen-demo` | `demo_launch-9a.d` beside `bin-demo-launch`, the relation `test_a_hyphenated_unit_is_matched_by_its_file_name` encodes |
| **a same-named unit's placement** | `other-<hash>/lib-probe.json` (already asserted) | `fingerprint("other", …)` lands under `other-<hash>` |
| **the uplift relation** | one file shared with the profile binary (already asserted) | `link=True` links the unit's output into the profile directory |

A fourth package (`hyphen-demo`, hyphenated binary target and a space-named source) carries what the last two comparisons need. Its reads do not touch `probe`'s units, so the three existing cases' assertions are unchanged.

## Verification

Each comparison was checked to **bite** by reverting the rule it covers in place:

| reverted rule | failing case | the failure |
|---|---|---|
| `unit()` spells every input absolute | dep-info spelling | `[True, True] != [False, True]`, both spellings named — **while the hand suite stays 60 passed**, which is the blindness this exists to cover |
| `fingerprint()` writes no `.json` | fingerprint pair | the fixture's pair is `['bin-demo-launch']` (25 hand tests catch this one too) |
| the checker drops relative resolution | — | 5 hand tests fail where 0 failed before the fixture spelled them as cargo does |

With everything restored: the hand suite **60**, the integrity suite **105** (103 + 2), housekeeping **11**, the legacy importers **1**, the workspace **659 passed / 0 failed** on stable and on the pinned **1.85.0**, `cargo fmt --all --check` and strict `clippy` clean, and the record check on freshly built binaries by both toolchains unchanged at **113 named / 0 unnamed** and **40 / 0**. Runtime added is under a second (the scratch workspace has no dependencies, so `--offline` has nothing to fetch), and the module still **skips with the reason** where no cargo can be run — re-checked in the four configurations after the skip path moved into a shared helper.

## Residual

* The fixtures **must stay hand-built** for the many shapes a rule has to refuse that no cheap real build produces: malformed framing and marks, an undecodable record, a missing binary, alias builds, sibling workspaces, a configuration file cargo does not read. What is compared here is only what cargo can be made to write cheaply.
* Real cargo cannot cheaply produce some of the shapes those fixtures assert *about cargo* either (a copied rather than linked artifact, a stale dep-info), so they remain decisions the fixture makes.
* Nothing in production code, `wire.txt` or the checker's rules moves; the build-script read outside every closure package stays the reported limitation.


RepairYourTech:
## Measurement record — head `85956e90`

Head-driven end to end on this tree, and the two comparison cases ran **on the CI runner** (the `offline-validation` log names them: 105 tests, no skips), so the scratch workspaces were built by the runner's own cargo.

### The drift, found by comparison rather than reasoning

The fixture's `unit()` wrote **every input absolute** (`str(self.root / read)`). Cargo writes a package's own sources **relative to the directory it ran in** and its own generated paths absolute. Measured on this repo's real scratch build:

```
/tmp/….d: probe/src/main.rs /tmp/…/target/debug/build/probe-9f1fd5fe027fd7c0/out/generated.rs
          ^ relative         ^ absolute
```

The consequence was not cosmetic and was measured directly: the branch every real build's sources take (a relative token, resolved against the workspace) was reached by **no test of the hand suite** — dropping the checker's relative resolution failed **0 of its 60 tests** before this change, and **5** now.

### What was compared, and how each comparison bites

| shape | cargo's own bytes | reverted fixture rule | the failure |
|---|---|---|---|
| dep-info spelling | source relative, `out/` input absolute | `unit()` spells every input absolute | `[True, True] != [False, True]` — **hand suite stays 60 passed** |
| dep-info contents | `# env-dep:OUT_DIR=`, a bare `path:` rule, `hyphen-demo/src/a\\ b.rs` | — | — |
| fingerprint | `<package>-<hash>/bin-<target>` + `.json` | `fingerprint()` writes no `.json` | pair is `['bin-demo-launch']` (25 hand tests catch it too) |
| identity spellings | `hyphen_demo-<hash>.d` beside `bin-hyphen-demo` | — | — |
| same-named unit's placement | `other-<hash>/lib-probe.json` | — | — |
| uplift relation | one file shared with the profile binary | — | — |

Two rows per comparison are asserted to have the same shape rather than described as agreeing.

### Battery

Hand suite **60**, integrity **105**, housekeeping **11**, legacy importers **1**; workspace **659 passed / 0 failed** on stable and on the pinned **1.85.0**; `cargo fmt --all --check` and strict `clippy` clean; the record check on binaries freshly built by each toolchain unchanged at **113 named / 0 unnamed** and **40 / 0**. The module still **skips with the reason** where no cargo can be run — re-checked in four configurations after the skip path moved into a shared helper (an empty `PATH` with `CARGO` unset, and with `CARGO=/nonexistent`, both report `OK (skipped=2)` rather than erroring).

### Corrected during this pass

* My first expectation string spelled the hyphenated package `crates/hyphen-demo/…`; the scratch workspace places it at `hyphen-demo/`.
* My first version of the dep-info comparison asserted **one** source token and **one** generated token; rustc writes each in **several** of the dep-info's rules, so both assertions were wrong and the suite caught it. The comparison now asserts the *spelling* is single and the tokens are as cargo wrote them.
* The skip check I ran first was invalid — `/usr/bin/cargo` exists here, so those "no cargo" environments were real builds, not skips. Redone with a genuinely cargo-less `PATH`.

### Residual

The fixtures **stay hand-built** for the many shapes no cheap real build produces (malformed framing and marks, an undecodable record, a missing binary, alias builds, sibling workspaces, a configuration file cargo does not read), and for shapes cargo cannot cheaply be made to produce *about itself* (a copied rather than linked artifact, a stale dep-info). Production code, `wire.txt` and the checker's rules do not move; the build-script read outside every closure package stays the reported limitation.

